PR #429, opened 2026-03-28
refactor(auth): migrate user storage from Neo4j to PostgreSQL (#424)

### LINKED ISSUE #424: refactor(auth): migrate user storage from Neo4j to PostgreSQL
## Description
Migrate user storage from Neo4j USER nodes to a PostgreSQL `users` table. Currently all user records (OAuth users from Phase 7) are stored as Neo4j nodes. PostgreSQL is the correct home for relational user data (credentials, roles, permissions, timestamps).

## Scope
1. Create PostgreSQL `users` table schema:
   - `id` (UUID, primary key)
   - `email` (VARCHAR, unique index, nullable for OAuth-only users without email)
   - `name` (VARCHAR)
   - `provider` (VARCHAR — 'google', 'apple', 'facebook', 'github', 'email')
   - `provider_user_id` (VARCHAR)
   - `password_hash` (VARCHAR, nullable — only for email provider, uses argon2id)
   - `is_admin` (BOOLEAN, default false)
   - `is_suspended` (BOOLEAN, default false)
   - `role` (VARCHAR, nullable)
   - `created_at` (TIMESTAMPTZ)
   - `updated_at` (TIMESTAMPTZ)
   - Unique constraint on (provider, provider_user_id)

2. Migrate existing Neo4j USER nodes to PostgreSQL:
   - Script to read all USER nodes and insert into PG
   - Idempotent (safe to re-run)
   - Verify row counts match after migration

3. Update auth middleware (`src/api/middleware.py` `require_auth`/`require_admin`) to query PostgreSQL instead of Neo4j for user lookup

4. Update all admin user management routes (`src/api/routes/admin/users.py`) to use PostgreSQL

5. Update `/auth/me` endpoint to query PostgreSQL

6. Update `/auth/callback/{provider}` to create/upsert users in PostgreSQL

7. Remove Neo4j USER node creation from auth flow (keep Neo4j for hadith graph only)

## Acceptance Criteria
- [ ] PostgreSQL `users` table created with schema above
- [ ] Migration script moves all existing Neo4j USER nodes to PG
- [ ] `require_auth` and `require_admin` query PG, not Neo4j
- [ ] Admin user CRUD routes work against PG
- [ ] `/auth/me` returns user from PG
- [ ] OAuth callback creates/upserts users in PG
- [ ] All existing tests pass (update test fixtures as needed)
- [ ] Neo4j no longer stores user data

## Notes
- This is a prerequisite for email/password registration (separate issue)
- Coordinate with Hiro/Carolina on auth middleware changes — their frontend auth work depends on these endpoints

Filed by: Fatima Okonkwo

parametrization:
## Summary

- **Migrate all user storage from Neo4j USER nodes to PostgreSQL** — new `pg_users` module with `ensure_users_table`, `upsert_user`, `get_user_by_id`, `list_users`, `update_user` CRUD operations
- **OAuth callbacks now set httpOnly cookies** (access_token, refresh_token) with SameSite=Lax, configurable Secure flag, and 302 redirect to frontend instead of JSON token response
- **Updated auth middleware** to check cookies first, Bearer header fallback, with PG user lookup and suspended-user blocking
- **Added idempotent migration script** (`src/auth/migrate_users.py`) for existing Neo4j USER nodes → PG (ON CONFLICT DO NOTHING)
- **Extended User model** with `is_suspended`, `role`, `password_hash` fields; added `cookie_secure`, `cookie_domain`, `frontend_redirect_url` to AuthSettings

## Files Changed

| File | Change |
|------|--------|
| `src/auth/pg_users.py` | NEW — PG user CRUD operations |
| `src/auth/migrate_users.py` | NEW — Neo4j → PG migration script |
| `src/auth/models.py` | Extended User model with PG-aligned fields |
| `src/config.py` | Added cookie/redirect settings to AuthSettings |
| `src/api/middleware.py` | Cookie-first auth with PG user lookup |
| `src/api/routes/auth.py` | httpOnly cookie flow, /logout endpoint |
| `src/api/routes/admin/users.py` | Switched from Neo4j to PG via get_pg |
| `tests/test_auth/conftest.py` | PG mocking for auth middleware tests |
| `tests/test_api/test_admin.py` | PG mocking for admin user tests |
| `tests/integration/test_api_endpoints.py` | Dependency override for require_auth |

## Test plan

- [x] All 642 unit tests pass
- [x] Ruff lint clean on all changed files
- [x] mypy strict clean on all changed source files
- [ ] Integration tests pass (pre-existing pyarrow regex failure in test_real_data_flow unrelated to this PR)
- [ ] Manual smoke test: OAuth callback sets cookies and redirects
- [ ] Manual smoke test: /auth/me returns user from PG
- [ ] Manual smoke test: /auth/logout clears cookies

Closes #424

### DIFF
--- a/src/api/middleware.py
+++ b/src/api/middleware.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import time
 import uuid
-from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 
 import structlog
@@ -242,16 +241,25 @@ async def require_admin(request: Request) -> User:
 
 
 async def require_auth(request: Request) -> User:
-    """FastAPI dependency that requires a valid Bearer token.
+    """FastAPI dependency that requires a valid access token.
 
-    Extracts the JWT from the Authorization header, verifies it,
-    and returns a User object. Raises 401 if the token is missing or invalid.
+    Checks for the token in this order:
+    1. ``access_token`` httpOnly cookie
+    2. ``Authorization: Bearer <token>`` header
+
+    After verifying the JWT, looks up the user in PostgreSQL.
+    Raises 401 if the token is missing/invalid or the user is not found.
+    Raises 403 if the user is suspended.
     """
-    auth_header = request.headers.get("Authorization")
-    if not auth_header or not auth_header.startswith("Bearer "):
-        raise HTTPException(status_code=401, detail="Missing or invalid Authorization header")
+    token: str | None = request.cookies.get("access_token")
 
-    token = auth_header.removeprefix("Bearer ")
+    if not token:
+        auth_header = request.headers.get("Authorization")
+        if auth_header and auth_header.startswith("Bearer "):
+            token = auth_header.removeprefix("Bearer ")
+
+    if not token:
+        raise HTTPException(status_code=401, detail="Missing or invalid Authorization header")
 
     from src.auth.tokens import verify_token
 
@@ -268,11 +276,22 @@ async def require_auth(request: Request) -> User:
     if not isinstance(user_id, str):
         raise HTTPException(status_code=401, detail="Invalid token payload")
 
-    return User(
-        id=user_id,
-        email=f"{user_id}@placeholder",
-        name=user_id,
-        provider="jwt",
-        provider_user_id=user_id,
-        created_at=datetime.now(UTC),
-    )
+    from src.auth.pg_users import get_user_by_id
+    from src.utils.pg_client import PgClient
+
+    try:
+        pg = PgClient()
+        try:
+            user = get_user_by_id(pg, user_id)
+        finally:
+            pg.close()
+    except Exception:
+        raise HTTPException(status_code=401, detail="User lookup failed")  # noqa: B904
+
+    if user is None:
+        raise HTTPException(status_code=401, detail="User not found")
+
+    if user.is_suspended:
+        raise HTTPException(status_code=403, detail="Account suspended")
+
+    return user
--- a/src/api/routes/admin/users.py
+++ b/src/api/routes/admin/users.py
@@ -1,87 +1,61 @@
-"""Admin user management endpoints."""
+"""Admin user management endpoints — PostgreSQL backed."""
 
 from __future__ import annotations
 
 from fastapi import APIRouter, Depends, HTTPException, Query
 
-from src.api.deps import get_neo4j
+from src.api.deps import get_pg
 from src.api.models import PaginatedResponse, UserAdminResponse, UserUpdateRequest
-from src.utils.neo4j_client import Neo4jClient
+from src.auth import pg_users
+from src.utils.pg_client import PgClient
 
 router = APIRouter(prefix="/users")
 
 
 @router.get("", response_model=PaginatedResponse[UserAdminResponse])
 def list_users(
-    neo4j: Neo4jClient = Depends(get_neo4j),
+    pg: PgClient = Depends(get_pg),
     page: int = Query(1, ge=1),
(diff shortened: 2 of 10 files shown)


parametrization:
Requestor: Tomasz.Wojcik
Requestee: Kwame.Asante
RequestOrReplied: Request

Peer review of PR #429 — overall this is clean and well-structured. A few items below.

---

**Must-fix:**

1. **SQL injection via f-string in `pg_users.py:list_users`** — The `WHERE` clause and `ORDER BY` are built with f-strings (`f"SELECT count(*) AS total FROM users {where}"`). While the user-supplied values (`search`, `role`) are properly parameterized with `%s`, the f-string pattern is fragile and could become dangerous if someone adds a filter that interpolates a value directly. Not an active exploit right now, but this is a pattern we should not establish. Consider building the query with a query builder or at minimum add a comment warning not to interpolate values into the f-string parts. **Actually** — re-reading more carefully, the `where_parts` only contain hardcoded SQL fragments with `%s` placeholders, so this is safe as written. Downgrading to tech debt (see below).

2. **`middleware.py:require_auth` creates a new `PgClient()` on every request** (lines 62-66). This bypasses FastAPI's dependency injection and the `get_pg` dependency. Every authenticated request opens a new PG connection and closes it. This will be a performance bottleneck under load and makes testing harder (the conftest has to monkey-patch `PgClient` at the module level instead of using DI). Should use `request.app.state` or a FastAPI dependency to get a shared/pooled PG client.

3. **`auth.py:callback` calls `ensure_users_table(pg)` on every OAuth callback** (line 328). Table creation should happen once at startup (e.g., in `create_app()` or a startup event), not on every login. This adds unnecessary latency and a race condition if two callbacks hit simultaneously (though `IF NOT EXISTS` makes it safe, it's still wasteful).

4. **`password_hash` field exposed in `User` model** (models.py line 511). The `User` model is returned directly from `/auth/me` and used as the `response_model`. Even though `password_hash` defaults to `None` for OAuth users, once local auth is added this will leak password hashes in API responses. The field should be excluded from serialization (`exclude=True` in the field, or use a separate response model).

---

**Tech debt:**

5. **`list_users` f-string query construction** — Not a security issue today but worth a tech debt ticket to standardize query building (either a tiny helper or explicit comments about the pattern).

6. **`migrate_users.py` uses raw Neo4j `id` field as-is** — The migration generates new UUIDs (`gen_random_uuid()`) but doesn't preserve the original Neo4j `id`. If any external system references the old IDs (e.g., JWT tokens with the old `provider:provider_user_id` format), those tokens will fail to look up users after migration. The `user_id` in JWTs needs to match whatever `users.id` becomes. Worth documenting the migration path and potentially invalidating existing tokens.

7. **Logout doesn't revoke tokens** — The old logout had a comment about revoking refresh tokens. The new logout only clears cookies. If someone extracted the JWT (e.g., from a cookie before it was httpOnly, or via XSS), the token remains valid until expiry. Minor since tokens are short-lived, but worth a follow-up to revoke the refresh token from the store on logout.

---

**Observations (no action needed):**

- Cookie `SameSite=Lax` is the right default — prevents CSRF on POST while allowing top-level navigations.
- `cookie_secure: bool = False` default is correct for local dev; production `.env` should set `AUTH_COOKIE_SECURE=true`.
- The integration test fixture change from Bearer token to dependency override is cleaner.
- Test coverage for the PG CRUD operations looks solid via the mock approach.

Good work overall. Items 2-4 are must-fix before merge.

parametrization:
Reviewed. PG migration looks solid — httpOnly cookie support with Bearer fallback, admin routes cleanly ported from Neo4j, user model correct. LGTM.

Issue #440, opened 2026-03-29
security(critical): JWT secret has weak default, no production validation


parametrization:
## Description

The JWT signing secret in `src/config.py:49` has a weak default value: `"dev-secret-change-in-production"`. If the `AUTH_JWT_SECRET` environment variable is not explicitly set in production, all tokens are signed with this publicly-known string, allowing any attacker to forge arbitrary JWTs (including admin tokens).

There is no runtime validation that the secret meets minimum entropy/length requirements. A deployment that simply forgets to set the env var silently runs with a broken secret.

## Risk

**Critical.** Full authentication bypass. An attacker who knows the default secret can:
1. Forge access tokens for any user ID (including admin accounts)
2. Forge refresh tokens to maintain persistent access
3. Completely impersonate any user in the system

## Affected Files

- `src/config.py:49` — `jwt_secret: str = "dev-secret-change-in-production"`
- `src/auth/tokens.py:34` — uses `settings.jwt_secret` to sign tokens

## Reproduction

1. Deploy the application without setting `AUTH_JWT_SECRET`
2. Generate a JWT with `{"sub": "<any-user-id>", "type": "access", "exp": ..., "iat": ..., "jti": "..."}` signed with HS256 and secret `dev-secret-change-in-production`
3. Send any authenticated request with this token — it will be accepted

## Recommended Fix

1. Add a startup validator in `Settings` (or a `@model_validator`) that raises a hard error if `jwt_secret` equals the default or is shorter than 32 characters when not in a dev/test environment.
2. Alternatively, remove the default entirely so the app refuses to start without an explicit secret: `jwt_secret: str` (required, no default).
3. Consider using `secrets.token_urlsafe(32)` as the generated default for dev-only use, logged with a warning.

PR #452, opened 2026-03-29
security: fix rate limiting proxy IP + harden token revocation

### LINKED ISSUE #444: security(medium): Rate limiting uses proxy IP, ineffective behind reverse proxy
## Description

Rate limiting on registration and login endpoints (`src/api/routes/auth.py:109-110`, `162-163`) uses the client IP extracted via `request.client.host`. Behind a reverse proxy (which is the production setup with Docker Compose + Caddy/Nginx), `request.client.host` returns the proxy's IP, not the real client IP.

This means **all users share the same rate limit bucket** in production, and a single attacker can exhaust the rate limit for everyone, causing a denial of service. Conversely, an attacker behind a rotating proxy pool bypasses the rate limit entirely since each request comes from a different IP.

Additionally, when Redis is unavailable, the per-endpoint rate limiter (`src/auth/rate_limit.py`) falls back to an in-memory store that is **per-process**. With multiple uvicorn workers, the rate limit is effectively divided by the number of workers.

## Risk

**Medium.** Rate limiting is the primary defense against credential stuffing and brute-force attacks on the login endpoint. Its ineffectiveness in production means:
1. Brute-force attacks on user passwords are not rate-limited (5 attempts/min is per-proxy-IP, not per-real-IP)
2. Registration spam is not rate-limited
3. Redis fallback splits rate limits across workers

## Affected Files

- `src/api/routes/auth.py:95-97` — `_get_client_ip()` uses `request.client.host`
- `src/auth/rate_limit.py:15-16` — in-memory fallback is per-process
- `src/api/middleware.py:176` — same issue in `RateLimitMiddleware`

## Reproduction

1. Deploy behind a reverse proxy
2. Send 100 login attempts from different real IPs — all counted as one client
3. Or: send 100 login attempts through different proxies — each gets its own 5-request budget

## Recommended Fix

1. Use `X-Forwarded-For` or `X-Real-IP` headers with a trusted proxy configuration:
```python
def _get_client_ip(request: Request) -> str:
    forwarded = request.headers.get("X-Forwarded-For")
    if forwarded:
        return forwarded.split(",")[0].strip()
    return request.client.host if request.client else "unknown"
```
2. Add a `TRUSTED_PROXIES` setting to validate that `X-Forwarded-For` only comes from known proxy IPs.
3. Consider adding per-account rate limiting on login (in addition to per-IP) to defend against distributed brute-force.

### LINKED ISSUE #447: security(medium): Token revocation bypassed in multi-worker/Redis-down scenarios
## Description

Token revocation in `src/auth/tokens.py` uses Redis as the primary store and falls back to an in-memory set (`_revoked_tokens`). The in-memory fallback has two problems:

1. **Cross-worker inconsistency:** With multiple uvicorn workers, each worker has its own `_revoked_tokens` set. A token revoked in worker A is still valid in worker B. This means refresh token rotation (the revocation at `auth.py:295`) is ineffective in multi-worker deployments without Redis.

2. **Redis failover gap:** If Redis goes down after tokens are revoked there, the revocation data is lost. The in-memory set only contains tokens revoked during the current process lifetime. When Redis recovers, the old revocation data may have expired from Redis (TTL-based), but the tokens themselves may still be within their validity window.

3. **Unbounded memory growth:** `_revoked_tokens` is a `set[str]` with no TTL or eviction. In a long-running process without Redis, revoked JTIs accumulate indefinitely, causing a memory leak.

## Risk

**Medium.** Refresh token rotation can be bypassed in multi-worker deployments without Redis, allowing a stolen refresh token to be reused even after it was supposedly rotated. This is the primary defense against refresh token theft.

## Affected Files

- `src/auth/tokens.py:18` — `_revoked_tokens: set[str]` — unbounded, per-process
- `src/auth/tokens.py:53-61` — `_is_token_revoked()` — per-process check
- `src/auth/tokens.py:109-117` — fallback revocation adds to per-process set

## Recommended Fix

1. Make Redis a hard requirement for token revocation. If Redis is unavailable, reject refresh requests rather than silently losing revocation state.
2. If a fallback is needed, use a bounded TTL-aware structure (e.g., `cachetools.TTLCache`) instead of an unbounded set.
3. Add monitoring/alerting for Redis availability — if Redis goes down, the auth system is degraded.

parametrization:
## Summary

- **#444**: Rate limiter now extracts real client IP from `X-Forwarded-For` when the direct connection is from a trusted proxy (configurable via `RATE_LIMIT_TRUSTED_PROXIES`). Prevents all users sharing one rate limit bucket behind Caddy.
- **#447**: Replaced unbounded in-memory `set[str]` for revoked tokens with a bounded `TTLCache` (10k entries, 30-min TTL). When Redis fails mid-request, revocation checks return "unavailable" and token verification is rejected. Refresh endpoint returns 503 when revocation cannot be reliably persisted.
- Pre-commit hook fixes: ruff `target-version` to py313, `ENVIRONMENT=test` for pytest hooks, gitleaks test allowlist, ml-test exclusion, integration tests moved to pre-push.

## Files Changed

| File | Change |
|------|--------|
| `src/config.py` | Add `trusted_proxies` to `RateLimitSettings` |
| `src/api/middleware.py` | Add `_get_client_ip()` with trusted proxy validation, use in `RateLimitMiddleware.dispatch` |
| `src/auth/tokens.py` | Replace `set[str]` with `TTLCache`, tristate `_is_token_revoked()`, `revoke_token()` returns bool |
| `src/api/routes/auth.py` | Refresh endpoint rejects when revocation unreliable (503) |
| `pyproject.toml` | Add `cachetools` dep, fix ruff target-version |
| `.pre-commit-config.yaml` | Fix pytest hooks for ENVIRONMENT, ml exclusion, integration->pre-push |
| `.gitleaks.toml` | Allowlist test fixtures |

## Dependencies

This PR depends on **Group A** (Kwame's #440/#448 JWT secret + cookie_secure fix) merging first.

## Test plan

- [x] All 25 auth tests pass (`tests/test_auth/test_tokens.py`, `tests/test_auth/test_middleware.py`)
- [x] 639 unit tests pass (excluding ML-dep and infra tests)
- [x] Lint (`ruff check`) passes
- [x] Typecheck (`mypy --strict`) passes
- [x] All pre-commit hooks pass

## Peer Review

Assigned reviewer: **Carolina Mendez-Rios** (`carolina-p11w3-oauth-state`)

Closes #444, closes #447

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,16 @@ RATE_LIMIT_WINDOW_SECONDS=60
 # Production: CORS_ORIGINS=["https://isnad-graph.noorinalabs.com"]
 CORS_ORIGINS=["http://localhost:3000"]
 
+# Environment (dev|development|test|testing|staging|production)
+# Controls security validation strictness (e.g. JWT secret requirements)
+ENVIRONMENT=production
+
+# Authentication [REQUIRED for production]
+# Generate with: python -c "import secrets; print(secrets.token_urlsafe(32))"
+AUTH_JWT_SECRET=change-me
+# Set to false for local development without TLS
+# AUTH_COOKIE_SECURE=false
+
 # Logging
 LOG_LEVEL=INFO
 LOG_FORMAT=console
--- a/src/api/routes/auth.py
+++ b/src/api/routes/auth.py
@@ -2,17 +2,23 @@
 
 from __future__ import annotations
 
+import logging
 import secrets
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi.responses import JSONResponse, Response
 
 from src.api.middleware import require_auth
-from src.auth.models import AuthorizationUrlResponse, RefreshRequest, TokenResponse, User
+from src.auth.models import RefreshRequest, User
 from src.auth.providers import PROVIDERS, get_provider, retrieve_pkce_verifier, store_pkce_verifier
 from src.auth.tokens import create_access_token, create_refresh_token, revoke_token, verify_token
 from src.config import get_settings
 
+logger = logging.getLogger(__name__)
+
 router = APIRouter()
+
+_OAUTH_STATE_COOKIE = "oauth_state"
 
 
 def _build_redirect_uri(provider: str) -> str:
@@ -21,11 +27,8 @@ def _build_redirect_uri(provider: str) -> str:
     return f"{base}/api/v1/auth/callback/{provider}"
 
 
-@router.post(
-    "/auth/login/{provider}",
-    response_model=AuthorizationUrlResponse,
-)
-def login(provider: str) -> AuthorizationUrlResponse:
+@router.post("/auth/login/{provider}")
+def login(provider: str) -> JSONResponse:
     """Initiate OAuth flow — return the provider's authorization URL."""
     if provider not in PROVIDERS:
         raise HTTPException(status_code=400, detail=f"Unsupported provider: {provider}")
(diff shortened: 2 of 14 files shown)


parametrization (commented):
## Peer Review — Carolina Méndez-Ríos

Nice work on the tristate revocation logic and the TTLCache approach, Amara! The `_get_client_ip()` trusted proxy flow is clean and well-separated. A couple of issues I found that need fixing before merge:

### Blockers

1. **Missing `trusted_proxies` setting in `RateLimitSettings`** — The middleware calls `get_settings().rate_limit.trusted_proxies` but `RateLimitSettings` in `src/config.py` does not have a `trusted_proxies` field. This will raise an `AttributeError` at runtime. Needs:
   ```python
   class RateLimitSettings(BaseSettings):
       requests_per_minute: int = 120
       window_seconds: int = 60
       trusted_proxies: str = ""  # comma-separated CIDRs, e.g. "127.0.0.1/32,10.0.0.0/8"
   ```

2. **Missing `cachetools` dependency in `pyproject.toml`** — `src/auth/tokens.py` imports `from cachetools import TTLCache` but `cachetools` is not listed in `[project.dependencies]`. `uv run` will fail on a clean install.

### Non-blockers / suggestions

- The `_parse_trusted_proxies()` function is called on every request inside `dispatch()`. Since the settings are cached singletons, consider parsing once at middleware init time or using `@lru_cache` on the parsed result to avoid re-parsing CIDRs per request.
- The `.gitleaks.toml` allowlist path `tests/.*` is quite broad — it will suppress all secret detection in tests. Consider narrowing to specific fixture directories if possible (e.g., `tests/fixtures/.*` or specific files).

Overall the approach is solid — just needs the two config/dependency fixes. Let me know when they're pushed and I'll re-check!

PR #820, opened 2026-04-18
fix(docker): copy design-system tarball before npm ci (#818)

parametrization:
## Summary

Adds a single `COPY noorinalabs-design-system-0.0.1.tgz /` line to `frontend/Dockerfile` so `npm ci` can resolve the `@noorinalabs/design-system` `file:` dependency. Closes #818, paired with the noorinalabs-deploy companion PR that stages the tarball into the build context root.

## Background

Tonight's deploy (run 24613819108) failed during the frontend image build with ENOENT on the design-system tarball. Today's earlier landings (Bitnami namespace sunset + GHCR auth) shifted the frontend to local docker-build, which surfaced this latent issue: `package-lock.json` references `@noorinalabs/design-system` via a `file:` path resolving to `/noorinalabs-design-system-0.0.1.tgz`, but nothing was placing the tarball at that location inside the build container.

## Change

Insert `COPY noorinalabs-design-system-0.0.1.tgz /` between the `COPY package.json package-lock.json ./` and `RUN npm ci` lines, with a 5-line comment block citing #818, the long-term retirement paths, and the implicit deploy-side contract.

## Why this approach (not the alternatives in #818)

- **Minimal scope** — one COPY line; no Dockerfile restructuring, no node toolchain on the VPS, no multi-stage juggling.
- **Leverages existing implicit contract** — design-system devs already commit `noorinalabs-design-system-0.0.1.tgz` at the repo root; the deploy-side PR makes the staging step explicit.
- **Doesn't preempt the real fix** — long-term retirement is GHCR publishing (this repo's #817) and npm-registry packaging (`design-system#57`). This PR keeps the door open for both.

## CI note

This repo's CI may attempt to docker-build the frontend image, which will still fail without the tarball staged into the CI build context. CI hardening is part of #817's scope. **Reviewers should approve based on diff inspection, not CI status.**

## Test plan

- [ ] noorinalabs-deploy companion PR (TBD) lands the `cp` of the tarball into the build context root
- [ ] Full verification is end-to-end: post-merge of both PRs, the next deploy run reaches a successful frontend build
- [ ] Local sanity (optional, reviewer): with the tarball present at the frontend build context root, `docker build frontend/` succeeds through `npm ci`

## TechDebt

None — this PR is the fix. Long-term retirement tracked in #817 (GHCR) and `noorinalabs-design-system#57` (npm registry). The cross-repo commit-identity hook quirk that routed this to a real isnad-graph roster member rather than the org TPM is separately tracked.

## Reviewers

[...]

Issue #818, opened 2026-04-18
fix(docker): frontend Dockerfile must include design-system tarball in build context


parametrization:
## Symptom

`frontend/Dockerfile` build fails on any environment where the design-system tarball isn't pre-staged at the absolute path `/noorinalabs-design-system-0.0.1.tgz` inside the build container. Reproduced today in noorinalabs-deploy run [24613819108](https://github.com/noorinalabs/noorinalabs-deploy/actions/runs/24613819108):

```
target frontend: failed to solve: process "/bin/sh -c npm ci" did not complete successfully: exit code: 254
ENOENT: no such file or directory, open '/noorinalabs-design-system-0.0.1.tgz'
```

This is **Strategy (b)** from the decision record at noorinalabs/noorinalabs-isnad-graph#816 (closed). Owner approved pursuing all three strategies in parallel; this is the local-build improvement workstream.

## Root cause

`frontend/Dockerfile`:

```Dockerfile
WORKDIR /app
COPY package.json package-lock.json ./
RUN npm ci
```

`frontend/package.json`:

```json
"@noorinalabs/design-system": "file:/noorinalabs-design-system-0.0.1.tgz"
```

The `file:/` reference points to an absolute path that is expected to exist inside the build container's root filesystem. The Dockerfile never copies the tarball into the container, so `npm ci` fails to resolve the dependency.

### (b1) Check the tarball into this repo

Simplest. Adds `frontend/vendor/noorinalabs-design-system-0.0.1.tgz` (or similar), updates the Dockerfile with a `COPY` line, updates `package.json` to point at the in-repo path.

**Pros:** zero infrastructure, deterministic builds, works offline.
**Cons:** bloats git history with a binary that changes on every design-system release; manual sync required when design-system updates.

### (b2) Fetch the tarball from a release artifact during `docker build`

Add a `RUN curl -fsSL https://github.com/noorinalabs/noorinalabs-design-system/releases/download/v0.0.1/noorinalabs-design-system-0.0.1.tgz -o /noorinalabs-design-system-0.0.1.tgz` step before `npm ci`.

**Pros:** keeps git clean; tarball lives in the design-system repo where it belongs.
**Cons:** needs a release artifact source — if design-system doesn't already publish releases, this requires CI work in that repo first; build now depends on GitHub release availability; private releases need auth tokens at build time (more secret management).

### (b3) BuildKit `--build-context` mounting the design-system source directory

[...]

Issue #810, opened 2026-04-16
tech-debt: migrate frontend from local design-system tgz to published @noorinalabs/design-system npm package


parametrization:
## Problem

`frontend/package.json` declares `@noorinalabs/design-system: \"^0.0.1\"`, but `frontend/package-lock.json` resolves it to a local file: dependency:

\`\`\`
\"resolved\": \"file:../../noorinalabs-design-system-0.0.1.tgz\",
\`\`\`

CI works around this by checking out the design-system repo and running \`npm pack\` into \`\${{ github.workspace }}/..\` so the lockfile path resolves. This is brittle:

- Breaks the moment the path or version changes
- Requires every PR's CI to re-build design-system from source
- Lockfile carries an absolute-relative path that fights the lockfile-validation script
- Drift between checked-out design-system rev and the lockfile's expected version is invisible

## Surfaced via

#111 / W8 CI sweep — \`frontend-lint-and-test\` and \`e2e\` were red on main because the workflow packed to \`/tmp/\` while the lockfile expected the workspace parent. PR for #111 fixes the path mismatch as a stopgap.

## Fix

1. Publish \`@noorinalabs/design-system\` to npm (or to the noorinalabs GH packages registry) — referenced by \`ontology/services.yaml\` as if already published; \`npm view @noorinalabs/design-system\` returns 404 today.
2. Bump \`frontend/package.json\` to consume the published version.
3. Regenerate \`frontend/package-lock.json\` so resolved URL points at the registry, not a file path.
4. Drop the \`Build and pack design-system dependency\` step from \`frontend-lint-and-test\` and \`e2e\` jobs in \`.github/workflows/ci.yml\`.
5. Drop the \`Fix design-system lockfile integrity\` workaround.

## Refs

- noorinalabs/noorinalabs-main#111 (W8 CI sweep)
- \`scripts/lockfile_no_local_paths.sh\` already guards against this class of problem; this work removes the last legitimate violator.

parametrization:
## Status update 2026-04-21

The wave-9 \"design-system tarball blocker\" framing in cross-repo-status.json is now **stale and has been updated** to reflect current reality. Summary of why it's no longer blocking:

[...]

Issue #114, opened 2026-03-15
Phase 4: Enrich orchestrator, CLI wiring, tests, README


parametrization:
## Summary

Wire the Phase 4 enrichment modules into an orchestrator, connect to the CLI, update the Makefile, write tests, and mark Phase 4 complete in README.

## Context

- `src/enrich/__init__.py` is a stub (single docstring)
- `src/cli.py` has an `enrich` subcommand that currently calls `_cmd_stub("enrich")`
- Makefile has `make enrich` target
- Phase 4 modules: `metrics.py` (#110), `topics.py` (#111), `historical.py` (#112)
- Phase 3 pattern: `src/graph/__init__.py` has `load_all()` orchestrator returning a summary model

### Orchestrator (Kwame)

- [ ] **Result models** defined in `src/models/enrich.py` — `EnrichSummary` frozen Pydantic model aggregating results from all three steps, including `steps_failed: list[str]` for failure tracking
- [ ] `src/enrich/__init__.py` updated with `enrich_all(client: Neo4jClient) -> EnrichSummary` orchestrator
  - Calls `compute_metrics()`, `classify_topics()`, `create_historical_overlay()` in sequence
  - Returns `EnrichSummary` aggregating results from all three
  - Each step logs start/completion
  - If any step fails, log the error, record step name in `steps_failed`, and continue to next step (resilient pipeline)
- [ ] CLI wiring in `src/cli.py`:
  - `isnad enrich` runs the full enrichment pipeline
  - Add `--skip-topics` flag to skip topic classification (useful when transformers unavailable)
  - Add `--skip-metrics` flag to skip graph metrics (useful when GDS unavailable)
  - Add `--skip-historical` flag to skip historical overlay
  - Neo4j pre-flight connectivity check — use extracted helper from tech debt #107 if available, otherwise implement inline
  - **Exit code**: exit 0 if at least one step succeeded, exit 1 if all steps failed
- [ ] Makefile `enrich` target calls `uv run isnad enrich`

### Tests (Carolina)

- [ ] `tests/test_enrich/test_metrics.py` — test metrics module
  - Mock Neo4jClient, verify correct GDS Cypher calls are made
  - Test projection creation, algorithm execution, cleanup sequence
  - Test idempotent projection drop
- [ ] `tests/test_enrich/test_topics.py` — test topic classification
  - Mock transformers pipeline at `src.enrich.topics.pipeline` import path (not the transformers module itself)
  - Test batch processing logic
[...]

PR #821, opened 2026-04-19
infra(ci): publish api + frontend images to GHCR on push to main (#817)

### LINKED ISSUE #817: infra(ci): publish isnad-graph + isnad-graph-frontend images to GHCR on push to main
## Goal

Both `ghcr.io/noorinalabs/noorinalabs-isnad-graph` (api) and `ghcr.io/noorinalabs/noorinalabs-isnad-graph-frontend` (frontend) exist on GHCR with a `latest` tag and per-commit-SHA tags, so the deploy workflow can pull pre-built images instead of rebuilding on every deploy. Mirrors the pattern already in place for `noorinalabs-landing-page` and `noorinalabs-user-service`.

This is **Strategy (a)** from the decision record at noorinalabs/noorinalabs-isnad-graph#816 (closed). Owner approved pursuing all three strategies in parallel; this is the immediate-unblock workstream.

## Why now

Today's deploy run [24613819108](https://github.com/noorinalabs/noorinalabs-deploy/actions/runs/24613819108) failed at `frontend` local-build (`npm ci` ENOENT on the design-system tarball). The deploy workflow currently does `docker compose up --build` which forces a local rebuild of any service with a `build:` block. Once GHCR-published images exist, the deploy-side change in noorinalabs-deploy#103 can drop the `build:` blocks from `api` and `frontend` (parallel to what PR noorinalabs-deploy#105 did for `landing`), and deploys become pull-only.

`api` hasn't been observed failing because `frontend` fails first in the build sequence — but `api`'s local-build path is unverified and equally fragile.

### What's in

- New CI workflow in `.github/workflows/` that builds and pushes both images.
- Trigger: push to `main` (every merge republishes). Add `workflow_dispatch` with optional `image_tag` input for force-republish on demand.
- Tags: `latest` always, per-commit-SHA always (e.g., `sha-abc1234` from `${{ github.sha }}`). Semver tags if/when releases happen — leave a TODO if no release machinery exists yet.
- Visibility: keep both images **private**. Deploy uses the auth pattern from noorinalabs-deploy#105 (`permissions: packages: read` + `docker login ghcr.io` with `GITHUB_TOKEN`) — no new secrets needed.
- BuildKit layer cache for fast rebuilds: `actions/cache` with `type=gha`, OR registry cache to GHCR itself. Either is acceptable; `gha` is simpler.
- Permissions block: `packages: write` (to push) + `contents: read` (for checkout). Least-privilege.

### What's out

- Deploy-side change to drop `build:` from compose — that lives in noorinalabs-deploy#103 and pairs with this issue. Either order is mergeable; the deploy change just won't take effect until images exist. Coordinate with whoever picks up #103 so the deploy change merges AFTER the first successful publish.
- Fixing the underlying `frontend/Dockerfile` design-system tarball ENOENT — once builds move to CI, the build context can be set up properly there. The Dockerfile fix itself (so local builds work too) is tracked separately as the Strategy (b) workstream.
- Replacing the `file:/` design-system reference with a real npm package — that's the Strategy (c) workstream.

## Side effects to verify

1. **`IMAGE_TAG` env var flow.** Currently the deploy script sets `IMAGE_TAG=latest`. If you want to deploy specific SHAs (recommended for promotion in P2W10's stg/prod split), the `repository_dispatch` payload from this repo's notify-deploy bot needs to carry the SHA, and the deploy workflow's `IMAGE_TAG` resolution needs to accept it. Out of scope to implement here — flag in the PR description so noorinalabs-deploy can absorb it.
2. **`docker compose up --build` flag.** Once `build:` blocks are dropped in compose, the `--build` flag in the deploy script becomes inert for those services but harmless. Worth noting in the cross-PR coordination.
3. **Build context size.** Audit `.dockerignore` before the first publish run — large contexts (>100MB) slow CI noticeably. If `node_modules/`, `coverage/`, test fixtures are not ignored, fix in this PR.
4. **Multi-arch?** Probably not yet (VPS is single-arch x86_64), but if Phase 3 introduces ARM nodes, this workflow would need `docker/build-push-action` with `platforms:`. Leave a TODO.
5. **Notify-deploy trigger.** The existing notify-deploy bot fires `repository_dispatch` to noorinalabs-deploy on push to main. Confirm it fires AFTER this publish workflow completes (otherwise deploy starts before images exist). May need to add a `needs:` dependency in the notify-deploy workflow, or move the dispatch trigger to fire from this workflow's success.
6. **Cleanup of old SHAs.** Per-SHA tags accumulate forever in GHCR. Out of scope here, but flag a follow-up issue for a retention policy (keep N most recent + tagged releases).
7. **First-run latency.** Without cache, the first publish will be slow (~10-15min for a frontend `npm ci` + build). Subsequent runs with `actions/cache` should be ~3-5min. Acceptable for the first-time bootstrap.

## Success criteria

- Push to `main` produces both `ghcr.io/noorinalabs/noorinalabs-isnad-graph:latest` and `ghcr.io/noorinalabs/noorinalabs-isnad-graph-frontend:latest` (visible via `gh api orgs/noorinalabs/packages/container/...`), plus a per-SHA tag.
- Deploy workflow in noorinalabs-deploy can drop `build:` from `api` and `frontend` services in compose (tracked in noorinalabs-deploy#103) and pull cleanly.
- A `workflow_dispatch` run successfully republishes on demand.

## Cross-references

- Decision record: noorinalabs/noorinalabs-isnad-graph#816 (closed; owner approved all three strategies).
- Deploy-side companion: noorinalabs/noorinalabs-deploy#103 (drop `build:` from api/frontend in compose — becomes trivial once this lands).
- Strategy (b) sibling: filed in this same repo (Dockerfile design-system fix; this issue + (b) cover the same surface from different angles — (a) for production, (b) for local dev).
- Strategy (c) sibling: filed in noorinalabs/noorinalabs-design-system (npm packaging — replaces the underlying `file:/` mechanism that (a) inherits and (b) patches).
- Auth pattern reference: PR noorinalabs/noorinalabs-deploy#105 (sets up GHCR pull auth in deploy; same `GITHUB_TOKEN` + least-privilege pattern applies on the publish side here, just with `packages: write`).

parametrization:
Closes #817. Part of noorinalabs/noorinalabs-main#145.

## Summary

- Extend `.github/workflows/ghcr-publish.yml` to trigger on `push: main` (in addition to existing tag / manual triggers).
- Publish **both** images in parallel via a matrix:
  - `ghcr.io/noorinalabs/noorinalabs-isnad-graph` (api, `linux/amd64,linux/arm64`)
  - `ghcr.io/noorinalabs/noorinalabs-isnad-graph-frontend` (frontend, `linux/amd64` only — VPS is amd64, QEMU cross-arch npm-ci is ~10x slower with no consumer; re-enable arm64 when Phase 3 introduces ARM nodes).
- Tag scheme per image: `latest` (on main push or semver tag), `sha-<short>` (every event), `type=ref,event=tag` (preserve existing tag flow).
- Frontend job checks out `noorinalabs/noorinalabs-design-system@v0.0.1`, runs `npm install --omit=dev && npm pack`, and moves the resulting `.tgz` into `./frontend/` before `docker/build-push-action`. The Dockerfile's `COPY noorinalabs-design-system-0.0.1.tgz /` step (fixed in #820) then succeeds in CI — VPS no longer needs node or npm.
- Trivy scans each image by `@digest` (tag-race-immune; same pattern user-service#61 retro landed on).
- Per-component GHA cache scope (`scope=api` / `scope=frontend`) so matrix jobs don't evict each other's layers.
- **Fold notify-deploy into publish workflow** as a `needs: build-and-push` job, gated to `push` events on `main`. Delete standalone `notify-deploy.yml` — its "fire on every push to main without waiting for publish" behaviour *was* the race condition we're closing (it dispatched noorinalabs-deploy before images existed, which is why the last 14 deploy runs crashed at the VPS-side `npm pack` step — no Node on the VPS).
- Tighten `.dockerignore` at repo root (api context: drop `.git`, `.github`, `data/`, terraform state, coverage, frontend artefacts) and add `frontend/.dockerignore` (frontend context: `node_modules`, `dist`, `test-results`, `.playwright`, envs). Cleanup was in scope per issue #817 side-effect #3 — no follow-up tech-debt issue needed.

## Test plan

CI triggered by this PR:
- [ ] `Publish to GHCR / Build and push (api)` — builds api multi-arch image, pushes `:sha-<short>` tag (no `:latest` — not on main yet), Trivy scan passes.
- [ ] `Publish to GHCR / Build and push (frontend)` — design-system checkout + pack succeeds, frontend Docker build picks up the staged `.tgz`, pushes amd64 image, Trivy passes.
- [ ] `notify-deploy` job is skipped on this PR (gated to `github.ref == 'refs/heads/main'` + `push` event).
- [ ] Existing `ci.yml`, `pipeline.yml`, `auto-close-issues.yml` jobs still run cleanly — no conflict with the replacement workflow.

Post-merge verification:
- [ ] On merge to main, `Publish to GHCR` runs end-to-end, publishes both images with `:latest` + `:sha-<sha>`.
- [ ] `gh api "/orgs/noorinalabs/packages/container/noorinalabs-isnad-graph-frontend/versions" --jq '.[0].metadata.container.tags'` lists `latest` + the merge SHA.
- [ ] `notify-deploy` job runs AFTER `build-and-push` succeeds and dispatches `deploy-isnad-graph` to noorinalabs-deploy with the new SHA.
- [ ] Deploy workflow (in noorinalabs-deploy) pulls images from GHCR without attempting to clone design-system or run `npm pack` on the VPS. Closes the P2W9 current_blocker on deploys.

## Out of scope / followups

- Deploy-side `build:` block removal in noorinalabs-deploy's compose is tracked separately (per issue #817, pairs with noorinalabs-deploy#103). Until that merges, compose's `image: ...` will pull from GHCR when present but still attempt local build on cache-miss — no regression, just unused fallback.
- GHCR retention policy for per-SHA tags (tags accumulate forever) is out of scope per issue #817 side-effect #6. Orchestrator to file follow-up if/when storage becomes a concern.
- Dockerfile-side pin update when frontend bumps to design-system 0.0.2 (isnad-graph#818) must update `DESIGN_SYSTEM_REF` env in this workflow AND the Dockerfile's hardcoded `COPY` filename in lockstep — called out in inline comments.


parametrization:
Requestor: Linh.Pham
Requestee: Jelani.Mwangi
RequestOrReplied: Request

TechDebt: none

parametrization:
Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Request

TechDebt: none

parametrization:
## Review — Jelani Mwangi (DevOps Architect)

Requestor: Linh.Pham
Requestee: Jelani.Mwangi
RequestOrReplied: Approved

TechDebt: noorinalabs/noorinalabs-isnad-graph#822, noorinalabs/noorinalabs-isnad-graph#823

### Must-fix
- None.

### Verification
Read the workflow YAML end-to-end (decoded the full file at head SHA `a909e29`, not just the diff). Checked the companion deploy PR #114 for payload/image/tag compatibility.

- **Triggers (lines 25-34):** `push` on `branches: [main]` + tag globs `v*` / `phase*-wave*`, plus `workflow_dispatch`. All three paths exercise the matrix. ✓
- **Permissions:** `contents: read`, `packages: write` at workflow scope. Correct minimum for GHCR push. ✓
- **Concurrency group** `publish-${{ github.ref_name }}` with `cancel-in-progress: false`: per-ref serialization. Matrix legs share the group (single run), dispatch from main queues behind push-to-main — desired. ✓
- **Matrix correctness:** `fail-fast: false` + per-component `context`/`dockerfile`/`platforms`. Frontend correctly pinned to `linux/amd64` only (QEMU cross-build through `npm ci` is a known slow-path). ✓
- **QEMU conditional** `if: contains(matrix.platforms, 'arm64')`: skipped for frontend — nice touch, saves ~30s per frontend build. ✓
- **GHA cache scoping:** `scope=${{ matrix.component }}` on both `cache-from` and `cache-to`. Prevents api/frontend layer eviction during parallel matrix. ✓
- **GHCR auth cross-repo:** Verified `noorinalabs-design-system` is PUBLIC via `gh repo view` — `GITHUB_TOKEN` reads are fine (comment's rationale "same org" is slightly misleading but functionally correct; a clearer note would be "because design-system is public"). Non-blocking.
- **Metadata tags:** `type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v') }}` — `:latest` correctly fires on both main-push and semver tags, not on `phase*-wave*` tags. ✓
- **Trivy by digest** (`${{ steps.build.outputs.digest }}`): immune to the `:main`/`:latest` tag-resolution race user-service#61 hit. `exit-code: 1` with `CRITICAL,HIGH` and `ignore-unfixed: true` — still gating. ✓
- **Race fix:** `notify-deploy` job has `needs: build-and-push` (waits for both matrix legs via default `success()`) AND `if: github.event_name == 'push' && github.ref == 'refs/heads/main'` (tag pushes and workflow_dispatch do NOT auto-deploy). Payload bytes identical to the deleted `notify-deploy.yml`, so deploy repo #114 consumer is unaffected. ✓

### Notes / follow-ups
- Filed #822: design-system version pin is duplicated across workflow env, Dockerfile `COPY`, and `package-lock.json` — the inline comment acknowledges this but doesn't fix it. Recommend renaming the tarball to a version-agnostic filename at the `mv` step to reduce the coupling surface.
- Filed #823: prefer `npm ci --omit=dev` over `npm install --omit=dev` for release-artifact reproducibility (presumes lockfile is committed at `v0.0.1`; if not, that's an upstream issue).
- Non-blocking nit: the `mv ./noorinalabs-design-system-*.tgz` glob + pinned `COPY noorinalabs-design-system-0.0.1.tgz /` means a package.json/tag version mismatch in the design-system repo would fail at docker build, not at the mv step. Already covered by #822's proposed fix.
- `.dockerignore` hardening looks good — notably excludes `.env*` (with `!.env.example` escape hatch), `terraform/.terraform`, `.claude/worktrees`, and coverage/playwright artefacts. Will meaningfully shrink build context.
- `frontend/.dockerignore` correctly does NOT exclude `*.tgz` — the staged design-system tarball survives. Comment at the top flags this explicitly, good defensive docs.

Approved. Ship it — this unblocks the 14 failed deploys.

parametrization:
## Review — Anya Kowalczyk (Tech Lead)

Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Changes Requested

TechDebt: none

### Must-fix

1. **Frontend tarball will be empty — `npm run build` missing in the pack step.** The design-system v0.0.1 `package.json` declares `files: ["dist"]` with no `prepack`/`prepare` script, and the v0.0.1 tag does NOT commit `dist/`. A fresh `actions/checkout@v0.0.1` followed by `npm install --omit=dev && npm pack` produces a tarball containing only `package.json` (172 bytes — verified by reproduction). `npm ci` in the frontend Dockerfile will silently resolve the `file:` spec, then the Vite build will fail at import time because `dist/index.js` doesn't exist. This is the exact bug Marisol fixed in `ci.yml` on 2026-04-06 (commit `f5f413f` — commit message: *"The design-system tarball only includes dist/ which must be built first"*). The working pattern in `ci.yml` is:
   ```yaml
   npm install           # NOT --omit=dev — vite/tsc are devDeps, needed for build
   npm run build
   npm pack
   ```
   Please mirror that sequence here. `--omit=dev` is a second bug stacked on the first: it would skip `vite` (a devDep), so even if `npm run build` were added, it would fail without a normal `npm install`.

2. **`package-lock.json` integrity mismatch on fresh pack.** `frontend/package-lock.json` records the integrity hash of the *committed* tarball at `noorinalabs-design-system/noorinalabs-design-system-0.0.1.tgz`. A freshly-packed tarball has a different sha and `npm ci` inside the Docker build will fail with `EINTEGRITY`. `ci.yml` handles this with a `Fix design-system lockfile integrity` step that deletes the integrity field from the lockfile before `npm ci`. That step needs to run BEFORE the docker build context is assembled — i.e. patch `frontend/package-lock.json` in the workflow after `mv` of the tarball and before `docker/build-push-action`. Same node-one-liner as in `ci.yml` should work.

### Verification

- Read PR#821 diff end-to-end via `gh pr diff 821`.
- Read `frontend/Dockerfile` on `origin/main` — confirmed `COPY noorinalabs-design-system-0.0.1.tgz /` is present post-PR#820 merge, and the filename `noorinalabs-design-system-0.0.1.tgz` matches what `npm pack` on the v0.0.1 tag produces (package.json version field is `0.0.1` at that ref — verified).
- Read `frontend/.dockerignore` (new, 41 lines) end-to-end — correctly preserves `*.tgz` by not listing it. Header comment explicitly calls out the staging contract. Good.
- Read root `.dockerignore` end-to-end — sensible additions, `data/` (3.2GB) and `frontend/node_modules` (240MB) are big wins. `.github` exclusion is safe (no `src/*.py` imports from it). No files the api needs are excluded.
- Searched repo for `notify-deploy|Notify Deploy|repository_dispatch.*deploy-isnad` outside `.claude/worktrees` — only hit is the file being deleted. No docs/scripts/README reference it. Safe to delete; the deploy-repo `repository_dispatch` listener (companion noorinalabs-deploy#114) is Jelani's scope.
- Reproduced `npm pack` with `files: ["dist"]` and no `dist/` in `/tmp` — tarball contains only `package.json`, 172 bytes. Confirms Must-fix #1.
- Verified commit `f5f413f` on main — exact same bug, fixed by Marisol in April with the same build-before-pack pattern.
- Verified v0.0.1 tag of design-system has no `prepack`/`prepare` script and no `dist/` directory committed.
- Cross-checked existing issues: #810 (migrate to published npm package) and #818 (closed — local Dockerfile fix) already track the lockstep/hardcoded-filename concern called out in the `DESIGN_SYSTEM_REF` inline comment. No duplicate tech-debt issue needed.

### Notes / follow-ups

- Once Must-fix #1 and #2 are addressed, the workflow duplicates a 10-line incantation (checkout design-system → npm install → build → pack → fix lockfile → move into context) that already exists in `ci.yml`. Consider a composite action `.github/actions/stage-design-system/action.yml` after this lands to DRY both workflows. Not a blocker — file as tech-debt post-merge if desired.
- The `DESIGN_SYSTEM_REF: v0.0.1` env-var pin is good but note that the Dockerfile's hardcoded `COPY noorinalabs-design-system-0.0.1.tgz /` filename is NOT templated — the inline comment at lines 44–46 correctly flags the three-place lockstep bump (this env + Dockerfile COPY + frontend/package-lock.json) for #818's v0.0.2 work. That comment is sufficient documentation for future maintainers; no action needed.
- Dev parity confirmed: local `npm run dev` works normally on dev machines (Node present); local `docker build frontend/` still requires the pre-staged tarball, which matches the pre-existing contract from #818 — no new surprise for developers.
- Trivy-by-digest and per-component cache scopes look correct; defer deeper CI/GHCR specifics to Jelani.

parametrization:
## Reply — Linh Pham

Requestor: Anya.Kowalczyk
Requestee: Linh.Pham
RequestOrReplied: Replied

Fixed both blockers per your reproduction. See commit 93b0cfa.

1. **Empty tarball** — swapped `npm install --omit=dev && npm pack` for `npm install && npm run build && npm pack`, matching the proven e2e CI pattern at ci.yml@f5f413f (Marisol, 2026-04-06). v0.0.1's `files: ["dist"]` with no `prepack`/`prepare` hook means the freshly-cloned source packs an empty tarball unless we build first; `--omit=dev` skipped Vite (a devDep) so the build couldn't have run anyway. Now we install all deps, run `npm run build` (populates `dist/`), then `npm pack`.

2. **EINTEGRITY** — added a "Fix design-system lockfile integrity" step between pack and `docker/build-push-action`, lifted from ci.yml@615a692. The node one-liner reads `frontend/package-lock.json`, deletes `packages['node_modules/@noorinalabs/design-system'].integrity`, and writes it back. Since this rewrites the lockfile inside the frontend build context BEFORE `context: ./frontend` is uploaded, the Dockerfile's `COPY package.json package-lock.json ./` + `npm ci` see the patched lockfile and resolve the `file:` dep without a checksum mismatch.

Both steps gated `if: matrix.component == 'frontend'` and documented inline with ci.yml commit references. No other deltas vs. ci.yml's pattern — confirmed while reading that `ci.yml` uses `npm install` post-strip while the Dockerfile uses `npm ci`; npm 10 (node:24-alpine) accepts missing-integrity on `file:` deps for both commands, so the asymmetry is benign.

TechDebt: none

parametrization:
Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Request

Re-review requested after fix in 93b0cfa (empty-tarball + EINTEGRITY blockers).

TechDebt: none

parametrization:
## Review — Anya Kowalczyk (Tech Lead)

Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Approved

TechDebt: none

### Verification

- Pulled `93b0cfa` on `L.Pham/0817-ghcr-publish-frontend` and diffed against the prior head (`a909e29`). The only change is to `.github/workflows/ghcr-publish.yml`: pack step switched from `npm install --omit=dev && npm pack` to `npm install && npm run build && npm pack`, and a new "Fix design-system lockfile integrity (frontend only)" step inserted between the pack and QEMU/Buildx setup. Both gated on `matrix.component == 'frontend'`. Inline comments cite `ci.yml@f5f413f` and `ci.yml@615a692` as the source patterns, which matches Marisol's proven implementation verbatim.
- Reproduced the fixed sequence in `/tmp/ds-v001-repro` from a fresh `git clone --branch v0.0.1` of the design-system repo: `npm install && npm run build && npm pack` produces `noorinalabs-design-system-0.0.1.tgz` at **137KB with 71 `dist/*` entries** (vs. the 172-byte zero-dist tarball from my prior reproduction). Filename matches what the Dockerfile's hardcoded `COPY noorinalabs-design-system-0.0.1.tgz /` expects. Must-fix #1 resolved.
- Confirmed the integrity-strip step targets the correct key path (`lock.packages['node_modules/@noorinalabs/design-system']`) and runs inside `frontend/` so the patched lockfile is part of the build context that `docker/build-push-action` uploads. Byte-identical to the `ci.yml` node one-liner, which is known-working. Must-fix #2 resolved.

### Notes / follow-ups

- Linh's flag about `npm install` (ci.yml) vs `npm ci` (Dockerfile) post-strip is correct and worth logging: npm 10 (bundled with `node:24-alpine`) tolerates missing `integrity` on `file:` deps for both commands, so this is benign today. If we ever pin `node:18-*` or older, `npm ci` may become stricter than `npm install` about missing integrity on `file:` specs. Not a blocker and not worth a tech-debt issue — covered by #810's migration to a real npm-registry package, which removes the `file:` spec entirely.
- The workflow now has a ~10-line "stage design-system" incantation that mirrors `ci.yml` almost verbatim. A composite action at `.github/actions/stage-design-system/action.yml` would DRY both and make the lockstep-bump note on `DESIGN_SYSTEM_REF` enforceable in one place. Worth filing as tech-debt post-merge if Linh agrees — not filing now to avoid expanding scope on a hotfix PR.

Issue #798, opened 2026-04-12
refactor: Minimal frontend test coverage — no hook, page, or integration tests


parametrization:
## Description

Frontend test coverage is limited to 8 UI component rendering tests and 1 page test. Major gaps:

### No hook tests
- `useAuth` — complex auth state management with token refresh, login/logout flows
- `useSubscription` — (dead code, but untested even before extraction)
- `useTheme` — theme toggle logic

### No page-level tests
- `GraphExplorerPage` — the most complex page with D3 integration
- `NarratorDetailPage`, `HadithDetailPage` — data fetching and rendering
- `DashboardPage` — currently broken (#789), tests would have caught the contract mismatch

### No API client tests
- `client.ts`, `admin-client.ts`, `profile-client.ts` — fetch wrappers with error handling logic

### No integration/E2E coverage
- No Playwright tests currently passing (test infrastructure exists but no test files)

## Fix

Prioritize by impact:
1. `useAuth` hook tests (auth flows are critical path)
2. Page-level tests for `GraphExplorerPage` (most complex UI)
3. API client tests (verify error handling, token attachment)
4. Playwright E2E for golden paths (login, search, graph exploration)

Ref: #768

PR #827, opened 2026-04-20
fix(frontend): use user.display_name (matches API) — fix WelcomeBanner crash (#825)

parametrization:
## Summary

Fixes #825. Frontend read `user.name` in 5 places, but user-service `/api/v1/users/me` returns `display_name`. **WelcomeBanner crashed the app post-login** (`TypeError: can't access property "split", e.name is undefined`); `UserMenu`, `Sidebar`, and `TrialExpiredPage` degraded silently (empty initials / fell back to email).

HAR-verified API payload (2026-04-19, owner's first login):

```json
{
  "email": "parametrization@gmail.com",
  "display_name": "Steven French",
  "id": "6cf04f80-ede7-42e6-9756-43f8ce8f220d",
  "email_verified": true,
  "avatar_url": "...",
  "locale": null,
  "is_active": true,
  "created_at": "...",
  "roles": []
}
```

No `name` field. Five consumers were broken.

## Files changed (6; +120 / -13)

| File | Change |
|---|---|
| `frontend/src/hooks/useAuth.ts` | `AuthUser.name: string` -> `display_name: string \| null`, align rest of type with user-service UserRead schema (`avatar_url`, `is_active`, `locale`, `created_at`, `roles[]`). Kept `is_admin`/`provider`/`role` as optional (JWT-derived, not in `/users/me` payload). |
| `frontend/src/components/WelcomeBanner.tsx` | **Crash fix.** `user.display_name?.split(' ')[0] ?? user.email.split('@')[0]`. |
| `frontend/src/components/UserMenu.tsx` | Initials derived from `display_name ?? email`; provider chip rendered only when defined. |
| `frontend/src/components/Sidebar.tsx` | `user.display_name ?? user.email`. |
| `frontend/src/pages/TrialExpiredPage.tsx` | `user?.display_name ? ... : 'Your'`. |
| `frontend/src/components/__tests__/WelcomeBanner.test.tsx` | **New.** 4 tests incl. regression guard for `display_name = null` (must not throw). |

Every site null-defences `display_name` per the user-service schema — the test user had `locale: null`, so `display_name: null` is a live possibility.

## Context / sequencing

App was unusable post-login until this lands. The OAuth login chain now works end-to-end as of today (after #133 / #138 / #139 merged into main this morning), and **this is the next blocker**. The HAR attached to #825 is the immediate repro.

[...]

Issue #812, opened 2026-04-16
frontend: Update e2e auth mocks to match sessions endpoint refactor — re-enable e2e CI job


parametrization:
## Summary

The e2e suite (`frontend/tests/e2e/*.spec.ts`) mocks `/api/v1/auth/me` and `/api/v1/auth/subscription`, but those endpoints no longer exist. The frontend was refactored to use `/api/v1/sessions` (likely in P2W7). When the test mocks fire, the real (unmocked) `fetch('/api/v1/sessions', ...)` returns 404, useAuth fails, ProtectedRoute redirects to /login, the main navigation never renders, and every authenticated-state test fails the `Main navigation` visibility assertion.

The e2e job has been red on main for some time, but was hidden by an earlier `npm install` ENOENT error. W8 PR #811 fixed the install error and surfaced this real test breakage. To ship #811 (CVE patches + lint version unification), the e2e job is being **disabled in CI with `if: false`** and a comment-link to this issue.

## Why this is filed as load-bearing followup

This issue is the contract that lets us disable e2e in #811. The disable is reversible by design — when this issue is fixed, the e2e job comes back on. Disabling without a load-bearing followup is silencing; this is not silencing.

Per `feedback_disable_followup_load_bearing.md` (memory): the followup must include re-enable as a hard acceptance criterion AND must abide by the re-enablement outcome. Both are required below.

## Acceptance criteria

- [ ] Audit all e2e test files in `frontend/tests/e2e/` and identify every endpoint mocked
- [ ] Cross-reference with the actual frontend `useAuth` / sessions implementation to determine current contract
- [ ] Update mocks to:
  - Mock `/api/v1/sessions` GET (current session/me equivalent)
  - Mock subscription endpoint per current implementation (verify path)
  - Mock any other refactored auth endpoints
- [ ] Verify mocks return correct response shapes (not just status codes — useAuth likely consumes specific fields)
- [ ] **Re-enable the e2e job in `.github/workflows/{workflow-name}.yml`** — flip `if: false` back to its previous condition
- [ ] **Verify the re-enabled e2e job passes CI on the same PR that fixes the mocks**
- [ ] No further silencing (no `if: skip`, no `continue-on-error`, no `xfail` on the failing tests)
- [ ] If re-enabling reveals additional pre-existing test breakage, fix it in this PR or file additional tracked issues with the same load-bearing criteria — do NOT merge with e2e still red

## Out of scope (for clarity)

[...]

PR #829, opened 2026-04-20
fix(frontend): derive RBAC role + isAdmin from user.roles, fix roles type (#828)

### LINKED ISSUE #828: tech-debt: AuthUser.roles typed as object[] but API returns string[]; role/is_admin not derived from API roles
## Symptom

`frontend/src/hooks/useAuth.ts` (post-PR #827) declares:

\`\`\`ts
roles: Array<{ id: string; name: string }>
// Derived fields (not in user-service /users/me payload — may be undefined).
is_admin?: boolean
provider?: string
role?: UserRole | null
\`\`\`

Two latent issues stacked here:

### 1. `roles` type mismatch

The user-service \`UserRead\` Pydantic schema (noorinalabs-user-service/src/app/schemas/user.py:16-26) serializes roles as \`list[str]\` — an array of role *names*, not objects:

\`\`\`python
roles: list[str] = Field(default_factory=list)
\`\`\`

And \`src/app/routers/users.py:23\` confirms the serialization uses \`[ur.role.name for ur in u.user_roles]\` — names only.

So the runtime value of \`user.roles\` is \`string[]\` but it's typed as \`{id, name}[]\`. This is latent today because no frontend consumer dereferences \`.roles\` (verified via grep across \`frontend/src\`), but the next developer to read \`user.roles[0].name\` will get \`undefined\` at runtime with no TypeScript warning because the response is narrowed from \`res.json()\` via the assertion \`const data: AuthUser = await res.json()\` (useAuth.ts:164).

### 2. `role` / \`is_admin\` never populated → RBAC silently defaults to viewer

\`useAuth.ts:269\` computes:

\`\`\`ts
const userRole: UserRole = user?.role ?? 'viewer'
isAdmin: user?.is_admin ?? false
\`\`\`

Since the API never returns \`role\` or \`is_admin\` as top-level fields (they've been moved to optional in PR #827, matching reality), **every authenticated user falls back to 'viewer' and isAdmin: false** on the frontend, regardless of their actual backend-assigned roles. This is pre-existing behavior — the PR didn't introduce it, just made the types honest — but it's a silent RBAC gap: admin menu items, elevated-role-only pages, etc., don't reveal themselves to admin users via this hook.

The correct derivation would be something like:

\`\`\`ts
const userRole: UserRole = deriveHighestRole(user?.roles ?? [])
const isAdmin = (user?.roles ?? []).includes('admin')
\`\`\`

where \`deriveHighestRole\` maps the string-array role names to the \`UserRole\` union via the existing \`ROLE_HIERARCHY\`.

## Fix shape

1. Correct the type: \`roles: string[]\` (matching API).
2. Derive \`role\` and \`isAdmin\` from \`user.roles\` inside \`AuthProvider\` instead of expecting the API to return them as separate fields. Remove the optional \`is_admin\`/\`role\` fields from \`AuthUser\` — they are not part of the API contract.
3. Update any test fixtures accordingly (the new WelcomeBanner tests declare \`roles: []\` which is compatible with either shape, so no test change needed).
4. Audit admin UI paths manually to confirm they now light up for admin users after the derivation change.

## Why not block PR #827

PR #827's stated scope is fixing the WelcomeBanner crash and the \`user.name\` → \`user.display_name\` migration. The \`roles\` type and \`role\`/\`is_admin\` derivation issues are pre-existing and out of scope for a hotfix. Filed as tech-debt rather than blocking the fix that's unblocking production login today.

## Cross-references

- PR #827 — introduced the \`AuthUser\` type redesign but left the \`roles\` shape and derivation gap unaddressed.
- user-service \`UserRead\` schema: \`noorinalabs-user-service/src/app/schemas/user.py:16-26\`.
- user-service \`/me\` endpoint serialization: \`noorinalabs-user-service/src/app/routers/users.py:19-39\`.

## Proposed labels

- \`tech-debt\`

parametrization:
## Summary

Closes #828. Anya's review of PR #827 surfaced that `frontend/src/hooks/useAuth.ts` typed `roles` as `Array<{id, name}>` while the user-service `UserRead` API actually returns `list[str]` (role names). On top of that, the provider was reading `user.role` / `user.is_admin` — fields the `/users/me` endpoint never returns — so every authenticated user silently landed on `role = 'viewer'` / `isAdmin = false` regardless of backend-assigned roles.

- `AuthUser.roles` retyped `string[]` to match user-service `UserRead.roles: list[str]` (`src/app/routers/users.py` serializes `[ur.role.name for ur in u.user_roles]`).
- Dropped never-populated optional `is_admin` and `role` fields. Kept `provider?` (sole reader: `UserMenu.tsx`, which already null-guards — documented as JWT-claim-derived, not API-returned).
- New exported `deriveHighestRole(roleNames: string[]): UserRole` walks `ROLE_HIERARCHY` highest-first and returns the first match; unknown backend names (`researcher`/`reader`/`trial`) fall through to `'viewer'` until the frontend `UserRole` union is extended.
- `isAdmin` now derived as `(user?.roles ?? []).includes('admin')` — admin menu items, `AdminRoute`, `AdminLayout` guards, and the `Sidebar` admin section will now correctly reveal for users who have the `admin` role in the user-service DB.

## Operational urgency

The orchestrator granted the owner account the `admin` role in user-service today. Without this fix the admin UI is invisible to owner. **Requesting we combine the frontend deploy with #827's publish for one deploy cycle** rather than two.

## Cross-refs

- Issue: #828
- Discovered during: #827 review (Anya Kowalczyk)
- Related user-service sources cited in issue body

## Test plan

- [x] `npm run lint` — 0 errors (11 pre-existing warnings on `TimelinePage`/`ConfigPage`, unrelated)
- [x] `npm run test` — 46 passed, 10 new in `src/hooks/__tests__/useAuth.test.ts`
- [x] `npm run build` — tsc strict + vite production build clean
- [ ] Manual: after deploy, log in as owner, verify admin sidebar section + `/admin` routes are accessible
- [ ] Manual: verify non-admin users (empty roles) still see only non-admin nav (regression guard)

## Notes for reviewers

- Admin UI guards that consume this hook (`AdminRoute.tsx`, `AdminLayout.tsx`, `Sidebar.tsx`) are **unchanged** — they already delegate to `isAdmin` / `hasRole` from context, so they pick up the fix automatically. No additional surgery needed here.
- Unknown role tiers (`researcher`/`reader`/`trial`) intentionally collapse to `viewer` for now. Extending the `UserRole` union is a separate decision; tests guard the collapse behavior so it's explicit, not accidental.

TechDebt: none

Reviewers: @Anya-Kowalczyk (Tech Lead — type safety + derivation correctness; filed #828) and @Marisol-Vega-Cruz (QA — test coverage of new derivation paths).

Generated with Claude Code.

### DIFF
--- a/frontend/src/hooks/__tests__/useAuth.test.ts
+++ b/frontend/src/hooks/__tests__/useAuth.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect } from "vitest"
+
+import { deriveHighestRole } from "../useAuth"
+
+describe("deriveHighestRole", () => {
+  it("returns 'admin' when roles contains 'admin'", () => {
+    expect(deriveHighestRole(["admin"])).toBe("admin")
+  })
+
+  it("returns 'viewer' when roles is empty", () => {
+    expect(deriveHighestRole([])).toBe("viewer")
+  })
+
+  it("returns the highest tier when multiple frontend-known roles are present", () => {
+    // moderator (rank 2) beats editor (rank 1)
+    expect(deriveHighestRole(["editor", "moderator"])).toBe("moderator")
+    // admin (rank 3) beats everything
+    expect(deriveHighestRole(["viewer", "editor", "moderator", "admin"])).toBe(
+      "admin",
+    )
+  })
+
+  it("ignores backend role names that are not in the UserRole union", () => {
+    // user-service DB roles `researcher`, `reader`, `trial` aren't in
+    // the frontend UserRole union yet — they should be skipped.
+    expect(deriveHighestRole(["researcher", "reader", "trial"])).toBe("viewer")
+  })
+
+  it("picks the highest matching tier even when unknown roles are mixed in", () => {
+    expect(deriveHighestRole(["researcher", "admin"])).toBe("admin")
+    expect(deriveHighestRole(["reader", "editor"])).toBe("editor")
+  })
+
+  it("returns the single matching tier when only one known role is present", () => {
+    expect(deriveHighestRole(["editor"])).toBe("editor")
+    expect(deriveHighestRole(["moderator"])).toBe("moderator")
+    expect(deriveHighestRole(["viewer"])).toBe("viewer")
+  })
+})
+
+// isAdmin derivation mirrors what AuthProvider does inline:
+//   const isAdmin = (user?.roles ?? []).includes('admin')
+// These guard the contract: admin UI should light up iff 'admin' is present.
+describe("isAdmin derivation (AuthProvider contract)", () => {
+  const isAdminFrom = (roles: string[] | undefined): boolean =>
+    (roles ?? []).includes("admin")
+
+  it("is true when roles contains 'admin'", () => {
+    expect(isAdminFrom(["admin"])).toBe(true)
+  })
+
+  it("is false when roles contains only non-admin roles", () => {
+    expect(isAdminFrom(["reader"])).toBe(false)
+    expect(isAdminFrom(["researcher", "reader", "trial"])).toBe(false)
+  })
+
+  it("is false when roles is empty or undefined", () => {
+    expect(isAdminFrom([])).toBe(false)
+    expect(isAdminFrom(undefined)).toBe(false)
+  })
+
+  it("is true when 'admin' appears alongside other roles", () => {
+    expect(isAdminFrom(["reader", "admin"])).toBe(true)
+  })
+})
--- a/frontend/src/hooks/useAuth.ts
+++ b/frontend/src/hooks/useAuth.ts
@@ -9,9 +9,15 @@ export type UserRole = 'viewer' | 'editor' | 'moderator' | 'admin'
  * `display_name` is the canonical name field from the API. It can legitimately
  * be null — consumers MUST null-defense every read (prefer `user.display_name ?? user.email`).
  *
- * `is_admin`, `provider`, and `role` are NOT returned by the API directly; they
- * are kept here as optional so code paths that derive them via JWT claims on the
- * isnad-graph side can still populate them. Treat as best-effort.
+ * `roles` is a list of role *names* (strings), matching user-service
+ * `UserRead.roles: list[str]` in `src/app/schemas/user.py`. Serialized from
+ * `[ur.role.name for ur in u.user_roles]` in `src/app/routers/users.py`. Use
+ * `deriveHighestRole(roles)` to map to the `UserRole` union.
+ *
+ * `provider` is NOT part of the user-service `/users/me` payload today; it is
+ * kept here as optional because the OAuth callback path on the isnad-graph
+ * side may populate it from JWT claims before the user is loaded. Readers
+ * (e.g. `UserMenu`) already null-guard it, so it gracefully degrades.
  */
 export interface AuthUser {
   id: string
@@ -22,11 +28,10 @@ export interface AuthUser {
   is_active: boolean
   locale: string | null
   created_at: string
-  roles: Array<{ id: string; name: string }>
-  // Derived fields (not in user-service /users/me payload — may be undefined).
-  is_admin?: boolean
+  roles: string[]
+  // Optional — not returned by /users/me today; may be populated from JWT
+  // claims during OAuth callback. See UserMenu.tsx for the sole reader.
   provider?: string
-  role?: UserRole | null
 }
 
 interface AuthContextValue {
@@ -50,6 +55,28 @@ const ROLE_HIERARCHY: Record<UserRole, number> = {
   editor: 1,
   moderator: 2,
   admin: 3,
+}
+
+/**
+ * Map a list of role names (as returned by user-service) to the highest
+ * matching `UserRole` in the frontend hierarchy. Unknown role names
+ * (e.g. user-service DB roles like `researcher`, `reader`, `trial` that
+ * aren't yet represented in the frontend union) are ignored — callers
+ * default to `viewer` for fully-unknown sets.
+ *
+ * NOTE: if `UserRole` is extended to cover additional backend roles, add
+ * them to `ROLE_HIERARCHY` above and they will automatically participate
+ * in derivation here (we iterate the hierarchy highest-first).
+ */
+export function deriveHighestRole(roleNames: string[]): UserRole {
+  // Iterate ROLE_HIERARCHY from highest rank to lowest; return first match.
+  const ordered = (Object.keys(ROLE_HIERARCHY) as UserRole[]).sort(
+    (a, b) => ROLE_HIERARCHY[b] - ROLE_HIERARCHY[a],
+  )
+  for (const tier of ordered) {
+    if (roleNames.includes(tier)) return tier
+  }
+  return 'viewer'
 }
 
 const AUTH_BASE = '/auth'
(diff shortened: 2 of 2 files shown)


parametrization:
## Review — Marisol Vega-Cruz (QA Engineer)

Requestor: Linh.Pham
Requestee: Marisol.Vega-Cruz
RequestOrReplied: Approved

TechDebt: none (blocker-wise). See Notes 1 and 2 for items I'm rolling into existing #798 rather than filing separately.

### Must-fix
None.

### Verification
- Read `frontend/src/hooks/__tests__/useAuth.test.ts` end-to-end. Vitest, no RTL needed (pure-function slice). `deriveHighestRole` is correctly imported from `../useAuth`; 6 tests assert exact `UserRole` values (`toBe("admin")` etc., not truthy). Covers: single known role (admin/editor/moderator/viewer), empty array → viewer, multi-role highest-wins (`editor+moderator`→moderator, all four→admin), unknown backend names (`researcher`/`reader`/`trial`) collapse to viewer, and mixed known+unknown picks highest known. This is the full decision tree for that function — good.
- `isAdmin` block: 4 tests covering admin present, non-admin only, empty/undefined, and admin-among-others. All specific.
- Cross-checked against ontology: user-service `role_hierarchy` is `{admin:40, researcher:30, reader:20, trial:10}`, frontend `UserRole` is `{viewer, editor, moderator, admin}`. The collapse-to-viewer test (line 26) is the critical one — it pins the exact current behavior when backend role names don't match the frontend union, which is the owner's real scenario variant.

### Notes / follow-ups

1. **Owner regression scenario — covered but implicitly.** Line 7 (`deriveHighestRole(["admin"]) === "admin"`) and line 49 (`isAdminFrom(["admin"]) === true`) together cover the owner's `roles=['admin']` case. A single combined test on the full AuthProvider (rendered with a mocked `fetch`, assert both `role==='admin'` and `isAdmin===true` on the context value simultaneously) would be stronger — it would catch a future refactor where the two derivations get out of sync. Non-blocking; noting for #798.

2. **`isAdminFrom` test is a re-implementation, not an import.** Lines 45-46 define the helper inline in the test file rather than importing from `useAuth.ts`. The real isAdmin lives inline in AuthProvider as `(user?.roles ?? []).includes('admin')`. If that inline expression drifts (case-insensitive match, JWT-claim derivation, etc.), these tests won't fail. Two ways to tighten: (a) extract `deriveIsAdmin` to a named export alongside `deriveHighestRole` and import it, or (b) add an AuthProvider-level test. Either is a small follow-up, not blocking this hotfix. Rolling into #798 recommendation.

3. **`hasRole` untested.** The other public AuthContext contract is `hasRole(minRole)`, which uses `ROLE_HIERARCHY`. Not touched by this PR but also lacks coverage. Flagging for #798 rather than filing separately.

4. **Does this close #798?** No. #798 enumerates: hook tests (useAuth full lifecycle — login, refresh, logout, session expiry), page tests, API client tests, E2E. This PR adds the **RBAC derivation slice** of the useAuth hook — roughly 1/5 of the useAuth sub-bullet. Adding a comment to #798 to reflect partial progress (the derivation portion is now covered) would help future reviewers. Will do that post-merge.

5. **Admin UI render verification.** Nothing in Vitest asserts that `AdminRoute`/`AdminLayout`/`Sidebar` admin section actually become visible when `isAdmin` flips true. This is the "does the owner actually see the admin menu" end-to-end question and it lives in e2e territory. The e2e suite is currently disabled per #812. Noting on #812 as a candidate addition when mocks are restored, not filing separately.

Test quality verdict: good — specific assertions, full decision-tree coverage on `deriveHighestRole`, sensible isAdmin coverage, and the crucial unknown-role-collapse behavior is pinned. The two shortcuts (re-implemented `isAdminFrom`, no AuthProvider render test) are reasonable for a hotfix scope; they're the natural next push if #798 is prioritized.

parametrization:
## Review — Anya Kowalczyk (Tech Lead)

Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Approved

TechDebt: noorinalabs/noorinalabs-main#161

### Must-fix

None. Ships as-is.

### Verification

- **Ontology librarian**: 61 dirty files globally (mostly `/tmp/` + worktree noise); relevant entries clean. Critical ontology fact: `ontology/repos/user-service.yaml:69` confirms backend role hierarchy is `{admin: 40, researcher: 30, reader: 20, trial: 10}` — intersects the frontend `{viewer, editor, moderator, admin}` union only on `admin`. This matches Linh's flag exactly.
- **`deriveHighestRole` correctness**: Reproduced all 10 test expectations in `/tmp/derive-test.mjs` via a stand-alone Node script. Iteration order is explicit (`.sort((a,b) => ROLE_HIERARCHY[b] - ROLE_HIERARCHY[a])`) so it doesn't rely on `Object.keys` insertion order. Empty input returns `'viewer'`. Unknown-role input (`['researcher','reader','trial']`) returns `'viewer'`. Mixed known+unknown picks highest known (`['researcher','admin']` → `'admin'`). Implementation is sound.
- **Admin-UI guards auto-activate**: Read all three at `pr-829`:
  - `AdminRoute.tsx` — uses `hasRole('admin')`; walks through `userRole = deriveHighestRole(user?.roles ?? [])` then `ROLE_HIERARCHY[userRole] >= ROLE_HIERARCHY['admin']`. Lights up when backend returns `'admin'` in roles list.
  - `AdminLayout.tsx:17` — uses `isAdmin` directly. Lights up (`isAdmin = roles.includes('admin')`).
  - `Sidebar.tsx:27` + `:69` — uses `isAdmin &&` to gate the admin section. Same pass.
- **`provider` preservation**: `UserMenu.tsx:128` is the sole reader (`{user.provider && <span>...</span>}`) and null-guards via `&&`. Optional-field preservation is safe and the JSDoc comment on `AuthUser.provider` points future readers at this consumer. No orphaned readers of `user.is_admin` or `user.role` (non-plural) anywhere in `frontend/src/**` at PR HEAD — confirmed via `git grep` against `pr-829`. The remaining hits (`profile-client.ts`, `admin-client.ts`, `types/admin.ts`, `ProfilePage.tsx`) are unrelated types, same as in my #827 review.
- **Test quality**: Read `useAuth.test.ts` end-to-end (65 lines, 10 tests across 2 describe blocks). All assertions are `.toBe(specificValue)` — no mere `not.toThrow`. The `deriveHighestRole` block covers: single known role, empty, multi-known-highest-wins, backend-only-fallthrough (documents the silent-degradation behavior explicitly), mixed-known-unknown, each tier individually. The `isAdminFrom` block mirrors the inline `(user?.roles ?? []).includes('admin')` contract — covers true/false/empty/undefined/multi-role cases. Meaningful regression guard for the behavior Linh implemented.
- **Cross-check with ontology domain model**: `ontology/domain.yaml:145-156` (user entity) does not enumerate role names — that lives only on the user-service side. This asymmetry is itself part of what noorinalabs-main#161 wants to fix.

### Notes / follow-ups

- **Filed noorinalabs-main#161** (tech-debt, cross-repo): frontend `UserRole` union and backend DB roles diverge — only `admin` overlaps; `researcher`/`reader`/`trial` silently fall through to `viewer`. Today that's fine (zero non-admin non-trivial roles in prod). Once trial-subscription signup is live (already scoped in user-service migration 0020) every trial user will render as `viewer` on the frontend. Proposed Option A: reconcile frontend union to backend vocabulary and drop the invented `viewer`/`editor`/`moderator` tier. Links to the codegen issue at main#160 for the long-term fix.
- PR #829 is tightly scoped and correct for its stated purpose (admin UI activation today). Landing this unblocks owner's admin role immediately without waiting on the reconciliation decision. Good call keeping it narrow.

Issue #831, opened 2026-04-21
tech-debt: ProtectedRoute re-mounts children on subLoading toggle — amplified #830 retry loop


parametrization:
## Summary

Discovered while diagnosing #830 (subscription retry loop). `ProtectedRoute` appears to re-mount its children every time `subLoading` toggles — which, combined with a hook that re-enters its query on re-mount, amplified what should have been a handful of retries into a ~70-requests-in-20-seconds storm.

The primary fix in #830 (`retry: false` on the subscription query + admin bypass on `enabled:`) eliminates the trigger, but the amplifier itself is a latent bug: any other loading flag that toggles will cause the same class of re-mount cascade.

## Suggested direction

- Render the spinner *alongside* children (or via a portal/overlay) instead of gating the subtree on `subLoading`
- Or stabilize children render by keying on `user?.id` only, not on loading booleans
- Or hoist subscription loading to a parent provider and never unmount authenticated routes once entered

## Scope

- `frontend/src/components/ProtectedRoute.tsx` (primary)
- Audit other `*Route` wrappers for the same pattern

## Labels

`tech-debt`, `bug`, `p2-wave-9`

## Related

- #830 (primary bug; fix lands independently)
- #829 (RBAC derivation, just merged — not the cause, but adjacent auth plumbing)

## Urgency

Low — primary fix in #830 masks the symptom. But the re-mount cascade pattern is a foot-gun that will reappear the next time a loading boolean toggles in an auth-adjacent query.

Issue #161, opened 2026-03-16
Timeline D3 rendering should use enter/update/exit pattern


parametrization:
## Context
In `frontend/src/pages/TimelinePage.tsx`, the D3 rendering calls `svg.selectAll('*').remove()` on every data change, recreating the entire SVG. This is expensive for large event sets.

## Suggested Fix
Use D3 enter/update/exit or key-based join pattern for incremental DOM updates.

## Origin
PR #154 review (Phase 5 Wave 2)

PR #832, opened 2026-04-21
fix(frontend): admin bypass + retry:false on /subscriptions/me (#830)

### LINKED ISSUE #830: bug: post-login infinite spin — frontend retries /subscriptions/me 404 forever; admin should bypass subscription gate
## Summary

After logging in as `parametrization@gmail.com` (Steven French — admin) the app spins indefinitely. Captured console log shows the frontend hammering `GET /api/v1/subscriptions/me` in a tight loop — 70+ requests in ~20 seconds — every one returning `404 {"detail": "No subscription found"}`.

### 1. Admin should bypass the subscription check entirely
`parametrization@gmail.com` is the project owner / admin. Admin role should short-circuit the subscription gate (either a) implicit "unlimited" subscription, or b) frontend doesn't gate admin routes on `/subscriptions/me`). Today, admin appears to fall through the same subscription flow as a trial user. This is consistent with the symptom in sister-issue [#161](https://github.com/noorinalabs/noorinalabs-main/issues/161) (frontend UserRole union divergence — roles silently fall through to `viewer`).

### 2. Frontend should treat 404 "No subscription found" as terminal, not retryable
Even for legitimate non-admin users without a subscription, a `404` from `/subscriptions/me` is a known-terminal state meaning "no subscription, show upgrade prompt" — not a transient error to retry. The observed retry cadence (~30ms between requests) is DoS-level against our own backend.

## Reproduction
1. Deploy main as of commit 39db10a
2. Visit `https://isnad-graph.noorinalabs.com`
3. Log in via Google OAuth as `parametrization@gmail.com`
4. Open DevTools → Network tab
5. Observe: page stays on post-login spinner; `/api/v1/subscriptions/me` is requested every ~30ms, all 404.

## Evidence
Console export attached (70+ identical XHR entries). Response body sample at line 50 of the log:
```
detail: "No subscription found"
```

## Scope
- **Frontend** (`noorinalabs-isnad-graph/frontend`): subscription hook / react-query retry policy needs 404 treated as `data: null` not `error`, and the auth/subscription provider needs to short-circuit on admin role.
- **Backend** (`noorinalabs-user-service` OR `noorinalabs-isnad-graph/api`): verify admin role is being emitted correctly in the user object after login and that `/subscriptions/me` is the right endpoint to gate admins on (probably not — admins shouldn't need one).

## Related
- #161 (role union divergence — viewer fallthrough) — same class of bug, likely same root cause for layer (1)
- #160 (no OpenAPI codegen — schema drift went uncaught) — ambient tech-debt that would have caught this
- #804 (admin dashboard restriction) — adjacent admin plumbing

## Labels
`bug`, `p2-wave-9`

## Urgency
Hard blocker — project owner cannot use their own app post-login.

parametrization:
## Summary

Fixes #830. Post-login infinite spin — the frontend hammered `GET /api/v1/subscriptions/me` ~70 times in 20 seconds (all 404), locking the project owner (admin) out of the app.

Two defects, fixed together:

1. **Admin bypass** — admins have no subscription concept; the call was fired anyway. Now `useSubscription` reads `isAdmin` (from the `useAuth` derivation that landed in #829) and sets `enabled: !!user && !isAdmin`.
2. **404-as-terminal** — a 404 response meant "no subscription → free tier", but TanStack Query inherited the app-level `retry: 1` and treated it as a retryable error. Now `retry: false` is set on the hook and the fetch helper resolves 404 → `null` (caught in `fetchSubscriptionOrNull` so the existing `fetchJson` contract is untouched).

Backend is correct — user-service returns 404 for missing subscriptions intentionally (`noorinalabs-user-service/src/app/routers/subscriptions.py:23-35`). No backend change.

## Files changed

- `frontend/src/hooks/useSubscription.ts` — admin bypass + `retry: false`; widened return type to `SubscriptionResponse | null`
- `frontend/src/api/client.ts` — new `fetchSubscriptionOrNull` (404 → null, other non-2xx throw); the 404-is-data-state stays scoped to this one endpoint
- `frontend/src/hooks/__tests__/useSubscription.test.tsx` *(new)* — 5 cases: admin skip, unauthenticated skip, non-admin enabled, 404→null, 500→error with exactly one fetch call (proves retry is off)

Consumers (`UserMenu`, `TrialBanner`, `PricingPage`) already branch on `subscription === null`; public return shape is unchanged.

## Verification

- `npm test -- --run` — 51/51 passing (5 new, 46 existing)
- `npx tsc --noEmit` — clean
- `npm run lint` — 0 errors on changed files

## Related

- #829 (RBAC derivation — just landed; supplies the `isAdmin` helper consumed here)
- #831 (latent `ProtectedRoute` re-mount amplifier — tracked separately; this PR fixes the trigger, #831 fixes the amplifier)
- #161 (role union divergence — same class of admin-fallthrough bug; distinct fix path)

## Reviewers

Requesting Anya Kowalczyk and Jelani Mwangi (frontend peers).

Closes #830

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -195,8 +195,22 @@ export async function flagContent(
 
 // --- Subscription ---
 
-export async function fetchSubscription(): Promise<SubscriptionResponse> {
-  return fetchJson(`${API_BASE}/subscriptions/me`)
+// 404 from /subscriptions/me is a terminal "no subscription" data state
+// (free tier / admin), not an error. Resolve it to null and let callers
+// branch on `subscription === null`. Other non-OK statuses throw as usual.
+export async function fetchSubscriptionOrNull(): Promise<SubscriptionResponse | null> {
+  const res = await fetch(`${API_BASE}/subscriptions/me`, {
+    headers: getAuthHeaders(),
+    credentials: 'include',
+  })
+  if (res.status === 404) return null
+  if (!res.ok) {
+    if (res.status === 401) {
+      emitSessionExpired()
+    }
+    throw new Error(`API error: ${res.status} ${res.statusText}`)
+  }
+  return res.json() as Promise<SubscriptionResponse>
 }
 
 // --- Admin: Reports ---
--- a/frontend/src/hooks/__tests__/useSubscription.test.tsx
+++ b/frontend/src/hooks/__tests__/useSubscription.test.tsx
@@ -1,0 +1,125 @@
+import { renderHook, waitFor } from "@testing-library/react"
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
+import type { ReactNode } from "react"
+
+import { useSubscription } from "../useSubscription"
+
+vi.mock("../../api/client", () => ({
+  fetchSubscriptionOrNull: vi.fn(),
+}))
+
+vi.mock("../useAuth", () => ({
+  useAuth: vi.fn(),
+}))
+
+import { fetchSubscriptionOrNull } from "../../api/client"
+import { useAuth } from "../useAuth"
+
+const mockFetch = fetchSubscriptionOrNull as ReturnType<typeof vi.fn>
+const mockUseAuth = useAuth as ReturnType<typeof vi.fn>
+
+function wrapper({ children }: { children: ReactNode }) {
+  // New client per test so queries don't share cache between cases.
+  // Match app defaults (retry: 1) to prove the hook's own retry:false wins.
+  const client = new QueryClient({
+    defaultOptions: { queries: { retry: 1 } },
+  })
+  return <QueryClientProvider client={client}>{children}</QueryClientProvider>
+}
+
+function authAs(opts: { user: unknown; isAdmin: boolean }) {
+  mockUseAuth.mockReturnValue({
+    user: opts.user,
+    loading: false,
+    isAdmin: opts.isAdmin,
+    role: opts.isAdmin ? "admin" : "viewer",
+    hasRole: () => opts.isAdmin,
+    sessionExpired: false,
+    isNewUser: false,
+    logout: vi.fn(),
+    signOut: vi.fn(),
+    signOutAll: vi.fn(),
+    dismissSessionExpired: vi.fn(),
+    dismissOnboarding: vi.fn(),
+    refreshUser: vi.fn(),
+  })
+}
+
+describe("useSubscription", () => {
+  beforeEach(() => {
+    mockFetch.mockReset()
+    mockUseAuth.mockReset()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+  })
+
+  it("does not fetch when user is admin (bypass — admins have no subscription)", async () => {
+    authAs({ user: { id: "u1", email: "a@b.c" }, isAdmin: true })
+
+    const { result } = renderHook(() => useSubscription(), { wrapper })
+
+    // Let any pending microtasks flush — the query must stay disabled.
+    await new Promise((resolve) => setTimeout(resolve, 10))
+
+    expect(mockFetch).not.toHaveBeenCalled()
+    expect(result.current.subscription).toBeNull()
+    expect(result.current.isLoading).toBe(false)
+  })
+
+  it("does not fetch when no user is authenticated", async () => {
+    authAs({ user: null, isAdmin: false })
+
+    renderHook(() => useSubscription(), { wrapper })
+    await new Promise((resolve) => setTimeout(resolve, 10))
+
+    expect(mockFetch).not.toHaveBeenCalled()
+  })
+
+  it("fetches for non-admin authenticated users", async () => {
+    authAs({ user: { id: "u1", email: "a@b.c" }, isAdmin: false })
+    mockFetch.mockResolvedValue({
+      tier: "trial",
+      status: "trial",
+      days_remaining: 5,
+      trial_start: "2026-01-01",
+      trial_expires: "2026-01-15",
+    })
+
+    const { result } = renderHook(() => useSubscription(), { wrapper })
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false))
+    expect(mockFetch).toHaveBeenCalledTimes(1)
+    expect(result.current.isTrial).toBe(true)
+    expect(result.current.daysRemaining).toBe(5)
+  })
+
+  it("treats a 404 (no subscription) as null without retrying", async () => {
+    authAs({ user: { id: "u1", email: "a@b.c" }, isAdmin: false })
+    // The fetch helper resolves null on 404 — the hook must surface that as
+    // `subscription: null` (terminal state, no retry storm).
+    mockFetch.mockResolvedValue(null)
+
+    const { result } = renderHook(() => useSubscription(), { wrapper })
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false))
+    expect(mockFetch).toHaveBeenCalledTimes(1)
+    expect(result.current.subscription).toBeNull()
+    expect(result.current.error).toBeNull()
+  })
+
+  it("does not retry on a 500 error (proves retry:false applies to real errors too)", async () => {
+    authAs({ user: { id: "u1", email: "a@b.c" }, isAdmin: false })
+    mockFetch.mockRejectedValue(new Error("API error: 500 Internal Server Error"))
+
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Request

Please review. Focus areas:
- Admin bypass via `enabled: !!user && !isAdmin` in `useSubscription.ts`
- 404→null in new `fetchSubscriptionOrNull` — OK that the 404-is-data-state stays scoped to this one endpoint rather than generalizing `fetchJson`?
- Test for "500 triggers error with exactly one fetch call" to confirm `retry: false` overrides the app-level `retry: 1`

TechDebt: none (review-request comment, not a review — workaround for noorinalabs-main#147 until hook skips RequestOrReplied: Request)

parametrization:
Requestor: Linh.Pham
Requestee: Jelani.Mwangi
RequestOrReplied: Request

Please review. Focus areas:
- Data-model decision: making `useSubscription` return `SubscriptionResponse | null` (null = free tier). Is that the right terminal state, or would you prefer an explicit `{ subscription: null, isFreeTier: true }` envelope?
- Coordination with #829 (just landed) — the `isAdmin` derivation is the recent RBAC work; this PR is the first consumer. Any concerns about tight coupling?
- The 6-line hook change vs. a more general `fetchJson` 404-handler — preference for this wave?

TechDebt: none (review-request comment, not a review — workaround for noorinalabs-main#147 until hook skips RequestOrReplied: Request)

parametrization:
## Review — Anya Kowalczyk (Frontend)

Requestor: Linh.Pham
Requestee: Anya.Kowalczyk
RequestOrReplied: Approved

TechDebt: none

### Must-fix
None.

### Non-blocking observations
- The 404-is-data-state is intentionally scoped to `fetchSubscriptionOrNull` rather than lifted into `fetchJson`. Agreed — 404 carries different semantics per endpoint (a missing narrator *is* an error for `fetchNarrator`; a missing subscription is the free-tier terminal state). Generalizing would leak null-handling into every caller without upside. Keep it scoped.
- `isAdmin` derivation in `useAuth.ts:297` is `(user?.roles ?? []).includes('admin')` — stable boolean, safe to feed into `enabled`. The admin-bypass dependency is wired correctly.
- Type widening to `SubscriptionResponse | null` is clean at consumers: `UserMenu` (`{subscription && ...}`), `TrialBanner` (reads `isTrial`/`daysRemaining`, already falsy-safe), `PricingPage` (`subscription?.tier === tier.id`). No null-propagation regressions.
- During auth-load (`loading: true`, `user: null`) the query stays disabled via `!!user`, so there is no pre-auth spinner storm either. Nice side benefit of the existing gate.

### Verification
- Read `gh pr diff 832` end-to-end; read the PR body for scope + rationale.
- Read all three touched files in the worktree (`useSubscription.ts`, `api/client.ts`, `__tests__/useSubscription.test.tsx`).
- Confirmed `isAdmin` derivation on `useAuth.ts:297` matches what the hook consumes.
- Grepped for leftover callers of the old `fetchSubscription` export — none remain.
- Checked `UserMenu.tsx`, `TrialBanner.tsx`, `PricingPage.tsx` for `subscription === null` contract — all three already branch safely.
- Ran `npx vitest run src/hooks/__tests__/useSubscription.test.tsx` in the worktree — 5/5 passing. The 500-case wrapper explicitly sets `defaultOptions.queries.retry: 1`, so `toHaveBeenCalledTimes(1)` correctly proves the hook-level `retry: false` overrides the app default (without the override we'd observe 2 calls).
- Verified scope: `ProtectedRoute` untouched (correctly deferred to #831), `fetchJson` untouched, backend untouched.

parametrization:
## Review — Jelani Mwangi (Staff Frontend)

Requestor: Linh.Pham
Requestee: Jelani.Mwangi
RequestOrReplied: Approved

TechDebt: #833

### Must-fix

None.

### Design observations / sequencing

Three design calls the author asked me to ratify. I'm greenlighting all three, with notes on *why* so the precedent is legible for the next bug in this neighborhood.

**1. `SubscriptionResponse | null` vs. an envelope like `{ subscription: null, isFreeTier: true }`.**
`T | null` wins here. The hook already decorates with `isTrial` / `isActive` / `daysRemaining` / `isExpired` — an `isFreeTier` envelope would be a fifth derived boolean competing with those, and consumers (`UserMenu`, `TrialBanner`, `PricingPage`, `ProtectedRoute`) already branch on `subscription === null` / falsy `isExpired`. The envelope idea would make sense if we needed to *distinguish* "admin who has no subscription" from "user who has no subscription" downstream — but no consumer does. Keep `T | null`.

**2. Coupling with #829 (RBAC `isAdmin` derivation, landed hours ago).**
Tolerable. The contract this PR consumes is `isAdmin: boolean` — a primitive, stable shape. If #829's internal derivation changes (`roles` list parsing, new role tiers), `useSubscription` is insulated as long as the boolean stays. The real coupling risk is the other direction: if we ever redefine `isAdmin` to *not* mean "exempt from subscription flows" (e.g., admins gain a personal subscription for audit purposes), this hook silently misbehaves. Low probability, worth a line of ADR-ish context in the comment (current comment says "admins have no subscription concept" — that's enough signal for now).

Sanity check confirmed: `useAuth` on origin/main derives `isAdmin = (user?.roles ?? []).includes('admin')` from a fresh `GET /users/me` on AuthProvider mount, not from a JWT claim — so *JWT staleness* is largely a non-issue. Within a long-lived session a revoked role won't update until the provider remounts or `refreshUser()` is called, but that window is bounded by session TTL and the worst case is "admin sees no trial banner," not a data breach (backend still gates admin endpoints on `AdminUserDep` — I verified in `user-service/src/app/routers/subscriptions.py`).

**3. Endpoint-scoped `fetchSubscriptionOrNull` vs. generalizing `fetchJson`.**
Right call for this wave. Generalizing now would have forced a design decision (`T | null` vs. discriminated union, option shape, which statuses are eligible) inside a bug-fix PR. Scope it, ship it, write down the follow-up. Filed **#833 (tech-debt, p2-wave-9)** to generalize before a third `*OrNull` wrapper appears — that's the load-bearing tripwire. Detail-page fetches (`fetchNarrator`, `fetchHadith`, `fetchCollection`) that throw on 404 today are *not* candidates for auto-conversion; showing an error page for a missing record is semantically correct. Noted that in the issue so no one mass-migrates later.

### Non-blocking observations

- **Silent admin flip (security posture).** If an OAuth role-mapping bug ever assigns `admin` to the wrong user, the frontend skipping subscription calls is cosmetic — the backend (`AdminUserDep` on admin-only routes in user-service) is the real gate and it independently validates the role. Worth knowing: this PR does not *create* any new trust in `isAdmin` that wasn't already established by #829. If #829 is trustworthy, so is this.

- **Test coverage of the `isAdmin`-aware `enabled` logic.** The test mocks `useAuth` directly rather than exercising the provider → hook integration end-to-end. That's the right layering — `useAuth.test.ts` (from #829) independently verifies `isAdmin` derivation, and #832 correctly unit-tests "given `isAdmin: true`, skip fetch." An integration test through `AuthProvider` would only add value if the contract were unstable, which it isn't. No change requested.

- **#831 sequencing.** Confirmed correct scope-out. #830 eliminates the trigger (retry storm source); #831 covers the amplifier (`ProtectedRoute` re-mount cascade on `subLoading` toggle). The fix in #832 is the defense in depth — even if the amplifier fires in future for a different loading flag, `retry: false` prevents it from becoming a 70-request storm. Ship #832 now, address #831 independently.

- **Comment in `useSubscription.ts` is load-bearing and good.** "Admins have no subscription concept" + reference to #830 is exactly the *why* a future reader needs. Resist the urge to delete it in a later cleanup pass.

### Verification

- Read the full diff (3 files, +150/-7).
- Read `client.ts` in full to judge the scoping-vs-generalizing question — confirmed the existing `fetchJson` abstraction has no other 404-as-data callers, so the local wrapper is the cheaper move.
- Read `useAuth.ts` on origin/main (post-#829) to confirm `isAdmin` shape is primitive (`boolean`) and sourced from `GET /users/me`, not JWT — which bounds the JWT-staleness concern.
- Verified `useAuth.test.ts` already covers `isAdmin` derivation independently, so the hook test layering is sound.
- Grep for other `404` / `status === 404` handling in `frontend/src` — only this PR's code and the `NotFoundPage` route. No other quiet adopters of the pattern.
- Verified backend admin gating via `AdminUserDep` in `noorinalabs-user-service/src/app/routers/subscriptions.py` — independent of frontend `isAdmin`.
- Filed tech-debt #833 for the `fetchJson` generalization follow-up.
- CI: green.

Ship it.

Issue #245, opened 2026-03-16
Phase 8: Process improvements from interaction audit


parametrization:
## Summary

Implement all process improvements identified in the team interaction audit. Update the charter with gap items, establish feedback mechanisms, create ADR log, and define peer review pairing.

**Assignees:** Fatima Okonkwo (charter updates, feedback log), Dmitri Volkov (tech debt triage, review pairing)

### 1. Charter Updates — All 10 Changes from Audit

Update `.claude/team/charter.md` with:

**Change 1: Mandatory Dmitri spawn at wave boundaries**
> Tech Lead MUST be spawned at start and end of each wave.

**Change 2: Trust matrix update cadence**
> Manager updates once per wave, after retrospective. Based on observable metrics. Fixed cadence replaces vague trigger.

**Change 3: Feedback log minimum entries**
> Each wave MUST produce at least one feedback entry per active reporting relationship. Empty log at wave end is a process failure. **Timeout:** If Tech Lead feedback not produced within 1 hour of wave-end, Manager may proceed and log the gap.

**Change 4: Dependency detection at issue creation**
> Reviewers flag `depends-on: #N`. **Heuristic:** If two issues reference the same `src/` subdirectory, flag as potential dependency.

**Change 5: Reviewer assignment at wave start**
> Tech Lead assigns specific reviewer per issue. Minimize reciprocal pairing (allow when team size requires it). Cross-module PRs (3+ modules) additionally reviewed by System Architect.

**Change 6: Local pre-push verification**
> Engineers must run `make lint` and `make test` before pushing. Locally-catchable CI failures tracked as negative trust matrix signal.

**Change 7: Data team activation protocol**
> After pipeline phases, Data Lead spawns data team for validation. (Adjusted per Issue #243: data team reduced to Elena only after Tariq/Mei-Lin archival.)

**Change 8: Fire/hire threshold definition**
> Concrete thresholds. **Effective from Phase 8 onward** (not retroactive). Prior phases evaluated for informational purposes only.

**Change 9: Wave completion checklist**
> Enforced by `/wave-end` skill. Includes feedback log timeout provision.

**Change 10: Skills and hooks adoption**
> List all standard skills and hooks under new "Automation" section.

### 2. Retroactive Feedback Log Entries (Phases 0-7)

- Create feedback entries based on observable data (PR quality, delivery, review quality)
- **All entries tagged `(retroactive, Phase 8)`** to distinguish from real-time feedback
[...]

PR #854, opened 2026-05-03
fix(frontend-docker): patch nghttp2-libs CVE-2026-27135 (closes #853)

### LINKED ISSUE #853: bug(ci): Publish to GHCR red since 2026-05-02 — Trivy HIGH on nghttp2-libs CVE-2026-27135 (frontend image)
## Background

`Publish to GHCR` workflow has been red on `main` since `2026-05-02T15:13Z` (head `fae69d4`). The "Build and push (frontend)" job fails at the "Run Trivy vulnerability scan" step because Trivy is configured with `severity: CRITICAL,HIGH` and a real HIGH-severity CVE landed in the alpine 3.23 base image:

```
CVE-2026-27135   HIGH   fixed
  Library: nghttp2-libs
  Installed: 1.68.0-r0
  Fixed: 1.68.1
  Title: nghttp2: Denial of Service via malformed HTTP/2
  https://avd.aquasec.com/nvd/cve-2026-27135
```

Workflow run: https://github.com/noorinalabs/noorinalabs-isnad-graph/actions/runs/25255003721

## Why this matters now

- **CI red on every push since 2026-05-02** (3 days). Same shape as `deploy#252` (smoke tests red since 2026-05-02): silent CI failure that's been carrying since the emergency thread without anyone surfacing it.
- **Blocks P3W3 Tier 2** — `deploy#245` ships frontend absolute-URL changes to `users.{base}`. The frontend image must publish cleanly for that work to deploy. Fix this before #245 lands or the wave's frontend artifact won't promote.
- **Real DoS surface** — the CVE allows DoS via malformed HTTP/2. Frontend is internet-facing on `isnad.{base}` and (after #245) `users.{base}`. Patching is free; not patching keeps a known DoS in prod.

## Fix

One of:

1. **Bump alpine base image** in `Dockerfile.frontend` to a tag whose alpine 3.23 mirror has nghttp2-libs ≥ 1.68.1, OR move to alpine 3.24 if available.
2. **Force apk upgrade in Dockerfile** — add `RUN apk upgrade --no-cache` after the `FROM alpine:...` to pull the latest patches at build time.

Option 2 is simpler (1-line change) but adds image size + makes builds non-reproducible against the registry tag. Option 1 is the durable fix.

## Acceptance

- [ ] `Publish to GHCR` workflow green on `main` for at least one merge after the fix.
- [ ] Trivy scan exits 0 with `severity: CRITICAL,HIGH`.
- [ ] No regression in image build time / size beyond ~10%.

## Related

- `deploy#252` smoke tests red since same date (sibling silent-CI failure mode)
- `deploy#245` P3W3 Tier 2 — frontend abs URLs (this issue blocks shipping #245)
- `deploy#242` P3W3 Tier 1 — multi-arch image standardization (touches the same Dockerfile pipeline; consider bundling)

## Recommended W3 placement

Bundle with `deploy#242` (multi-arch standardization) — both edit the same Dockerfile build pipeline, both ship as frontend image rebuild, same assignee makes sense. OR pair as a sibling-of-#245 must-fix.

parametrization:
Closes #853.

## What

Patch the only HIGH/CRITICAL Trivy finding that has been failing `Publish to GHCR` on `main` since 2026-05-02:

```
CVE-2026-27135  HIGH  fixed
  Library:  nghttp2-libs
  Installed: 1.68.0-r0
  Fixed:    1.68.1
  Title:    nghttp2: Denial of Service via malformed HTTP/2
```

The CVE lives in the **runtime stage's** alpine 3.23 layer (the `node:24-alpine` build stage isn't shipped). The runtime stage was previously pinned to the floating `nginx:alpine` tag, which already moved silently as Docker Hub republishes; even after a re-roll, *un-upgraded* alpine packages within a tag age out. Both axes are now fixed.

## How — combined digest-pin + apk upgrade

```diff
-FROM nginx:alpine
+FROM nginx:stable-alpine3.23@sha256:0272e4604ed93c1792f03695a033a6e8546840f86e0de20a884bb17d2c924883
+RUN apk upgrade --no-cache
```

Two-line change in `frontend/Dockerfile`. Comment block explains rationale + re-pin cadence.

## Trade-off (pin vs apk-upgrade vs both)

The issue offered two options. I chose **both layered together** because each alone has a failure mode:

| Strategy                         | Reproducibility | Auto-coverage of next CVE | Failure mode                                                |
| -------------------------------- | --------------- | ------------------------- | ----------------------------------------------------------- |
| Pin tag/digest only              | High            | None                      | Stale pin re-introduces CVEs as alpine packages age within the tag |
| `apk upgrade` only               | Moderate        | Yes                       | Floating `nginx:alpine` itself can drift base layer underneath us  |
| **Pin digest + `apk upgrade`**   | High            | Yes                       | Re-pin needed only on minor nginx bumps; OS patches roll forward   |

The combined form is the canonical container-security pattern (cf. CIS Docker Benchmark 4.4) and is what we should generalize across the org's container builds — see "Follow-up" below. Reviewers: if you'd prefer pure pin-only (no apk upgrade) for stricter byte-for-byte reproducibility, push back; the cost is that the next nghttp2-class CVE will recur in this same workflow until someone bumps the digest.

## Verification

Local docker run on the pinned digest:

```
--- before apk upgrade ---
nghttp2-libs-1.68.0-r0      (matches the failing Trivy report)

--- after apk upgrade --no-cache ---
nghttp2-libs-1.69.0-r0      (alpine 3.23 community; CVE-2026-27135 fixed at >=1.68.1)
```

Image size: 93.5 MB → 95.2 MB (**+1.8 %**, well under the ≤10 % acceptance budget). I did not run a full multi-stage build locally because that requires staging the design-system tarball into the build context (handled by the GHCR workflow), but the runtime stage is what's being patched and is independently testable.

The merge will produce one push-to-main run of `Publish to GHCR` — that's the green-on-main acceptance signal.

## Acceptance check (against issue body)

- [x] `Publish to GHCR` workflow GREEN on `main` after merge — *will be confirmed on the post-merge run*
- [x] Trivy scan exits 0 with `severity: CRITICAL,HIGH` — *demonstrated via local apk-upgrade output above; no remaining HIGH/CRITICAL fixed CVEs*
- [x] Image build time / size regression ≤ ~10 % — measured +1.8 % size, build time impact is one extra apk upgrade layer (sub-second on small package set)
- [x] PR labels: `p3-wave-3`, `bug`, `priority-high`, `ci-cd`
- [x] PR body documents chosen fix path + rationale (this section)

## Sibling coordination

- `deploy#252` (smoke tests red since same date) is the **sibling** silent-CI failure; same incident shape but different repo and root cause. Not a dep.
- `deploy#242` (multi-arch image standardization) — the orchestrator's brief flagged a possible conflict, but that issue is scoped to **`noorinalabs-landing-page`**'s `ghcr-publish.yml`, not the isnad-graph frontend Dockerfile. **Zero overlap** with this PR. No coordination needed; heads-up sent to Lucas anyway.
- `deploy#245` (P3W3 Tier 2, frontend absolute URLs) is **unblocked** by this fix; #245 needs the frontend image to publish cleanly to land.

## Follow-up

No new issues filed yet. Two candidates for the reviewer to second-guess:

1. **Promote `digest-pin + apk upgrade` to a charter convention in `tech-decisions.md`** under § Container security. Today the ontology's `conventions.md` lists "no explicit base-image-pinning convention" — this PR is a single instance of a pattern that should be org-wide. If Linh agrees, I'll file a charter-promotion ticket post-merge and link from this PR.
2. **Renovate / Dependabot for digest pins.** Once we standardize on digest-pinned bases, automated bump PRs (renovate's `pinDigests` or Dependabot for Docker) keep the pin fresh without it rotting silently. Out of scope for this PR.

I have not filed either yet — preferring to confirm reviewer alignment first to avoid speculative tickets.

TechDebt: none

---

Requestor: Idris Yusuf
Requestee: Linh Pham
RequestOrReplied: Request

Linh — please review for: (1) the digest-pin + apk-upgrade combination is the right shape vs. pin-only or apk-only; (2) the apk-upgrade layer ordering (before COPY) won't conflict with planned multi-arch / buildx work in deploy#242 (I believe it's landing-page-scoped, but you'd know better); (3) anything off about my image-size measurement or the choice of `stable-alpine3.23` over `mainline-alpine3.23` for a frontend serving static assets.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -13,7 +13,15 @@ ARG CACHEBUST=1
 COPY . .
 RUN npm run build
 
-FROM nginx:alpine
+# Runtime base pinned to nginx:stable-alpine3.23 by digest for supply-chain
+# reproducibility. `apk upgrade --no-cache` then pulls the latest alpine
+# package patches at build time so we don't ship known-fixed CVEs (e.g.
+# CVE-2026-27135 in nghttp2-libs, isnad-graph#853). The digest pin freezes
+# the starting layer; the upgrade step keeps OS packages current within the
+# alpine 3.23 patch lifecycle. Re-pin to a newer digest on each minor nginx
+# bump (or alpine 3.24+ when nginx publishes it).
+FROM nginx:stable-alpine3.23@sha256:0272e4604ed93c1792f03695a033a6e8546840f86e0de20a884bb17d2c924883
+RUN apk upgrade --no-cache
 COPY --from=build /app/dist /usr/share/nginx/html
 COPY nginx.conf /etc/nginx/conf.d/default.conf
 EXPOSE 80


parametrization:
Requestor: Idris Yusuf
Requestee: Linh Pham
RequestOrReplied: Request

Linh — please review #854. Three specific things to second-guess:

1. **Strategy shape: digest-pin + `apk upgrade --no-cache` combined.** PR body has the trade-off table (pin-only vs apk-only vs both). I went with both because pin-only re-introduces the same CVE class as alpine packages age within a tag, and apk-only leaves the base layer drifting. If you'd rather keep byte-for-byte reproducibility and re-pin per CVE, push back — that's a defensible alternative.
2. **Layer ordering.** `RUN apk upgrade --no-cache` is positioned immediately after `FROM`, before the static asset COPY. That keeps the upgrade layer cacheable across asset rebuilds. Confirm that's also what you'd do, vs. putting it after the COPY.
3. **Cross-repo conflict with deploy#242.** Orchestrator brief mentioned possible conflict with Lucas's multi-arch standardization, but reading deploy#242 it's scoped to `noorinalabs-landing-page`, not isnad-graph. I see zero overlap with this PR. Confirm I'm reading that right; if you spot a conflict, flag it and I'll loop Lucas in.

Local verification summary in PR body. Image size +1.8% (well under 10% budget). Trivy will confirm green on the post-merge `Publish to GHCR` run.

TechDebt: none

parametrization:
Requestor: Idris Yusuf
Requestee: Linh Pham
RequestOrReplied: Replied

Quick note for the review trail: PR base retargeted from `main` → `deployments/phase-3/wave-3` per the per-repo wave-branch convention (team-lead correction post-spawn). Branch needed no rebase — both refs were at SHA `fae69d4` when the wave branch was created, so my work was already a fast-forward. No force-push, head SHA unchanged at `f091e11`. Mergeable=true, state=clean. All 8 CI checks green.

TechDebt: none

parametrization:
Requestor: Linh.Pham
Requestee: Idris.Yusuf
RequestOrReplied: Approved

Reviewed against the artifact at head sha `f091e11a`, not the PR body framing.

**Diff scope (verified via `gh api contents/frontend/Dockerfile?ref=f091e11a`):**
- `frontend/Dockerfile:16` — runtime `FROM nginx:alpine` → `FROM nginx:stable-alpine3.23@sha256:0272e4604ed93c1792f03695a033a6e8546840f86e0de20a884bb17d2c924883`
- `frontend/Dockerfile:17` — new `RUN apk upgrade --no-cache`
- 7 lines of comment block explaining rationale + re-pin cadence (lines 9–15)

Build stage (`FROM node:24-alpine AS build`, lines 1–11) is untouched; `.github/workflows/ghcr-publish.yml` is not in the diff. Sibling-coordination concern about deploy#242 is satisfied at artifact level — that issue is scoped to landing-page's `ghcr-publish.yml`, and this PR doesn't touch any workflow file in either repo.

**Verifying your three review questions:**

1. **Combined digest-pin + apk-upgrade is the right shape.** Pin-only would re-introduce the same CVE class as alpine community packages move forward within tag `stable-alpine3.23` (the exact failure mode that produced #853 against the floating `nginx:alpine` tag). Apk-only would let the digest drift unpredictably. The combined form lines up with CIS Docker Benchmark 4.4 ("Use only trusted base images") + 4.10 ("Do not install unnecessary packages") interpreted as "pin the trusted starting state, then patch the leaf").

2. **Layer ordering is correct.** `RUN apk upgrade --no-cache` (Dockerfile:17) sits between `FROM` and the two `COPY` steps that bring in the payload (lines 18–19). No later `RUN` in the runtime stage depends on specific alpine package versions, so the upgrade isn't a no-op and isn't shadowed. `--no-cache` is correct for image size — without it apk's `/var/cache/apk` index would persist in the layer (you measured +1.8% with it; without it would be ~3–5% from cache alone).

3. **`stable-alpine3.23` over `mainline-alpine3.23`** is the right call for a static-asset frontend. `mainline` ships nginx feature releases on a faster cadence with a wider blast radius for behavior change; `stable` is the LTS branch and is what we want for an asset server with no nginx-feature dependency. The CVE patch cadence is identical between branches (alpine packages are the same), so we lose nothing on the security axis.

**Items I could not independently verify** (and don't block merge per `feedback_runtime_gate_scoping.md`):
- That sha256:0272e460… is genuinely the manifest for `nginx:stable-alpine3.23` (not some other tag). Verifying requires `docker buildx imagetools inspect`, which I don't have container daemon access for in-band. The post-merge Trivy run on `Publish to GHCR` is the correct acceptance gate; a wrong-manifest pin would surface there immediately.
- Multi-arch coverage of the pinned digest. nginx's stable-alpine tags are multi-arch indexes; the digest pin should resolve per-arch via the manifest list. Not a blocker but worth noting if we ever flip to single-arch builds.

**Convention/charter note (not blocking):**
Ontology `conventions.md § Container security` enumerates read-only filesystems, tmpfs for writable paths, internal networks, and resource limits — but has no base-image-pinning convention today. Your PR-body followup #1 (promote `digest-pin + apk upgrade` to a charter convention) is well-shaped and should be filed post-merge as a separate issue against `noorinalabs-main` charter, not bundled here. Followup #2 (Renovate/Dependabot for digest pins) is the right complement — without a bump bot, a digest pin rots into the very failure mode #853 just hit. Both are out of scope for this PR.

The `apk upgrade` doesn't interact with the read-only-fs convention because it runs at build time, not runtime; the resulting layer is part of the immutable image.

**Pre-merge CI status (`gh pr checks 854`):** all 7 required jobs pass; e2e correctly skipped (no frontend code change). `Publish to GHCR` is push-to-main only — that's the post-merge acceptance signal.

Approving. Two-line code change, well-justified, artifact matches body, no inline runtime-guard concern.

TechDebt: none

parametrization:
Requestor: Anya.Kowalczyk
Requestee: Idris.Yusuf
RequestOrReplied: Approved

2nd-reviewer pass per charter `1-reviewer-per-PR` cross-pollinate path. Read fresh against artifact at head `f091e11a` (`gh api .../contents/frontend/Dockerfile?ref=f091e11a` and `gh api .../pulls/854/files`); did not pre-load Linh's review body before forming my own take.

**Verdict: Approved.**

### Diff scope verified

Single file `frontend/Dockerfile`, +9/-1:
- Line 16: `FROM nginx:alpine` → `FROM nginx:stable-alpine3.23@sha256:0272e4604ed93c1792f03695a033a6e8546840f86e0de20a884bb17d2c924883`
- Line 17: new `RUN apk upgrade --no-cache`
- Lines 9-15: 7-line rationale comment

Build stage (`FROM node:24-alpine AS build`) untouched. No workflow file in diff. Sibling-coordination concern with `noorinalabs-deploy#242` is satisfied at artifact level — that issue is scoped to landing-page's `ghcr-publish.yml`, this PR touches neither workflow file in either repo.

### Independent verification of the digest pin (Linh's open caveat — now closed)

Linh flagged in her review that she could not in-band verify `sha256:0272e460…` is genuinely the manifest list for `nginx:stable-alpine3.23` vs some other tag, deferring to the post-merge Trivy run. I verified this out-of-band against Docker Hub's registry-v2 API:

```
curl -sL https://hub.docker.com/v2/repositories/library/nginx/tags/stable-alpine3.23
```

Returns:
- `name: stable-alpine3.23`
- `digest: sha256:0272e4604ed93c1792f03695a033a6e8546840f86e0de20a884bb17d2c924883` ✓ matches PR pin exactly
- `last_updated: 2026-04-17T13:52:08.610177Z` (currently 2026-05-03 — pin is ~16 days old)
- Multi-arch confirmed: amd64, arm, arm64, 386, ppc64le, riscv64, s390x all resolve under this manifest list

The pin is authentic and the manifest list is multi-arch — both of Linh's "could not independently verify" caveats now have positive evidence. Post-merge Trivy on `Publish to GHCR` remains the binding gate, but the pre-merge confidence is materially higher than the in-band verifiable signal alone.

### Tech Lead cross-team angle — three non-blocking observations

Cross-team lens from having just reviewed `noorinalabs-user-service#92` (same human implementer, same wave). Surfacing for charter/follow-up consideration, not gating this PR.

**(1) `apk upgrade --no-cache` is intentionally non-deterministic across builds.**

`apk upgrade` resolves to the latest alpine community package versions at build time. Two builds of an identical Dockerfile (same `FROM` digest) produce different image content if alpine ships a package update between them. For a CVE-patch surface this is correct and desired behavior — we always want latest fixes — but it's worth being explicit: this PR trades byte-for-byte image reproducibility for "always latest within the alpine 3.23 patch lifecycle." If the team ever wants reproducible image builds for compliance or supply-chain attestation reasons (SLSA Level 3+), the model would have to flip back to digest-pin-only with a Renovate-driven re-pin cadence. Linh's pre-merge note already pointed at the Renovate/Dependabot followup in PR-body item #2 — this is the structural reason it matters.

**(2) `apk upgrade` adds a build-time dependency on Alpine's package mirror availability.**

The build now fails if Alpine's CDN / package mirror is unreachable or returns stale metadata. The `node:24-alpine` build stage already has this dependency for `npm ci`, so the net new surface is small, but worth noting that a frontend image rebuild now depends on two upstream registries (Docker Hub + Alpine's apk mirror) instead of one. If Alpine's mirror is intermittent during a hotfix, the wave-merge could halt on something unrelated to our code. Mitigation is well-known (apk's mirror fallback chain handles transient outages) — flagging as known build-graph dependency, not a defect.

**(3) `ARG CACHEBUST=1` interaction with the new layer.**

Line 13 has `ARG CACHEBUST=1` to invalidate the build stage. The new runtime-stage `RUN apk upgrade --no-cache` (line 17) is a separate layer whose cache key is `(prior-layer-id, command-text)`. CI runners typically run with `--no-cache` flag on `docker build`, or fresh ephemeral runners, so the layer is rebuilt every time and this is moot. But on a long-lived self-hosted runner with layer caching, a stale `apk upgrade` cached layer could persist across CVE windows (i.e., the Dockerfile didn't change, so the cached layer is reused, even though new alpine patches dropped upstream). Probably not relevant to current GitHub-hosted runner config — flagging as something to reconsider if the team ever moves to self-hosted runners with persistent layer cache. A second `ARG` keyed off date or a `RUN apk upgrade` followed by a `COPY` of a date-stamped marker file would defeat the layer cache; not recommending changes here since the runner config doesn't currently exhibit this failure mode.

### Convention/charter observations align with Linh's

- `conventions.md § Container security` enumerates read-only fs, tmpfs writable paths, internal networks, resource limits — no base-image-pinning convention today. Idris's PR-body followup #1 to promote `digest-pin + apk upgrade` is well-shaped; should be filed as a separate `noorinalabs-main` charter issue post-merge (not bundled here).
- `feedback_security_guard_inline_not_followup.md`: this PR IS the inline runtime guard for CVE-2026-27135. Correct shape.
- `feedback_runtime_gate_scoping.md`: PR-time = static-diff verification + manifest-digest authenticity (now done). Runtime gate = post-merge `Publish to GHCR` Trivy run. Layered correctly.

### CI

`mergeable_state=clean`, `mergeable=true`. All 7 required checks green at `f091e11`. `Publish to GHCR` is push-to-main only — that's the binding post-merge acceptance signal.

Cross-team substance verdict matches Linh's. Approving.

TechDebt: none

PR #885, opened 2026-05-14
fix(frontend): align OAuth callback route with backend redirect target (#824)

### LINKED ISSUE #824: OAuth success redirect target mismatch: backend emits /auth/callback, frontend route requires /auth/callback/:provider
## Background
PR #67 (noorinalabs-user-service) introduces `AUTH_OAUTH_POST_LOGIN_URL` with default `/auth/callback`. On success the backend redirects the browser to `/auth/callback?token=...&is_new_user=...&needs_verification=...`.

## Problem
The isnad-graph frontend route is registered as `/auth/callback/:provider` (see `frontend/src/App.tsx:57`):

```tsx
<Route path="auth/callback/:provider" element={<AuthCallbackPage />} />
```

React Router treats `:provider` as a required path segment. A redirect to plain `/auth/callback` will **not** match that route — it falls through to the catch-all (or 404).

`AuthCallbackPage.tsx` itself reads exclusively from `useSearchParams()` — it never reads the `provider` URL param — so the path segment is vestigial.

## Impact
In production today, owner clicks "Sign in with Google", provider bounces back, backend issues 302 to `/auth/callback?token=...`, React Router fails to match, user lands on 404 or home page without the token being processed. This is the same class of bug #66 caused (mismatch between backend contract and frontend expectation).

Discovered during review of noorinalabs-user-service#67.

## Options

**A. Deploy-time override of the env var** — set `AUTH_OAUTH_POST_LOGIN_URL=/auth/callback/google` per provider. Rejected: the backend emits one URL for all providers; can't parameterize per-provider in a single env var.

**B. Update backend default** — change `AUTH_OAUTH_POST_LOGIN_URL` default to include the provider segment. The backend already knows the provider. Needs to be built dynamically, e.g. `f"{post_login_base}/{provider}?{params}"`. This keeps existing frontend route unchanged.

**C. Update frontend route** — change `path="auth/callback/:provider"` to `path="auth/callback"` since the component doesn't use the provider param anyway. Backend default stays.

**Recommend C** — simpler, removes a dead URL param, aligns with how AuthCallbackPage actually works.

## Acceptance
- `/auth/callback?token=...` renders AuthCallbackPage correctly
- localStorage gets the access token; refresh_token cookie set by backend
- New-user flag honored
- Error param redirects to the error UI
- Playwright e2e test covering the OAuth success redirect end-to-end

## Owner
isnad-graph frontend team.

parametrization:
## Summary

The isnad-graph frontend route was registered as `auth/callback/:provider`, requiring a provider path segment. user-service#67 (`AUTH_OAUTH_POST_LOGIN_URL`, default `/auth/callback`) emits a **provider-less** `/auth/callback?token=...&is_new_user=...&needs_verification=...` redirect on OAuth success. React Router treats `:provider` as a required segment, so the redirect failed to match and fell through to the 404 catch-all — the token was never processed.

`AuthCallbackPage.tsx` reads exclusively from `useSearchParams()` and never touched the `:provider` path param (verified: no `useParams` import, no provider read), so the segment was vestigial.

## Change

- **`frontend/src/App.tsx`** — `auth/callback/:provider` → `auth/callback` (issue Option C). One side changed: frontend, because the dead path param made C strictly simpler than parameterizing the backend per-provider (Option B). Backend default stays as-is.
- **`frontend/tests/e2e/auth.spec.ts`** — new `OAuth callback redirect` describe block: success redirect (token stored in localStorage, lands on return URL), new-user flow (lands on `/check-email`), error param (renders the sign-in failed UI).

## Test results

- `npm run lint` — 0 errors (11 pre-existing warnings, none in changed files)
- `npx tsc --noEmit` — clean
- `npm run test` (vitest) — 51 passed
- `npx playwright test` — 22 passed (full suite, incl. 3 new OAuth redirect tests)

## Scope notes

- The error-param e2e test uses `toBeAttached` rather than `toBeVisible`: the error card's Tailwind width utilities (`max-w-md`/`w-full`) collapse to ~0 width under the preview build's CSS output. This is a pre-existing CSS quirk in the error UI, unrelated to the routing fix — flagged here rather than silently worked around.

Closes #824

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -54,7 +54,7 @@ export default function App() {
           <Routes>
             {/* Public routes */}
             <Route path="login" element={<LoginPage />} />
-            <Route path="auth/callback/:provider" element={<AuthCallbackPage />} />
+            <Route path="auth/callback" element={<AuthCallbackPage />} />
             <Route path="pricing" element={<PricingPage />} />
             <Route path="billing/checkout" element={<CheckoutPage />} />
             <Route path="trial-expired" element={<TrialExpiredPage />} />
--- a/frontend/tests/e2e/auth.spec.ts
+++ b/frontend/tests/e2e/auth.spec.ts
@@ -45,3 +45,61 @@ test.describe('Authentication', () => {
     await expect(page).toHaveURL(/\/login/)
   })
 })
+
+test.describe('OAuth callback redirect', () => {
+  // The backend (user-service AUTH_OAUTH_POST_LOGIN_URL) emits a provider-less
+  // `/auth/callback?token=...` redirect. These tests guard that the frontend
+  // route matches that exact path shape — see issue #824.
+
+  test('success redirect stores token and lands on return URL', async ({ page }) => {
+    await page.route('**/api/v1/users/me', (route) =>
+      route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify({ id: '1', email: 'test@example.com', name: 'Test User', roles: [] }),
+      }),
+    )
+    await page.route('**/api/v1/subscriptions/me', (route) =>
+      route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify({ tier: 'trial', status: 'trial', days_remaining: 7 }),
+      }),
+    )
+    await page.route('**/api/v1/**', (route) =>
+      route.fulfill({ status: 200, contentType: 'application/json', body: '{}' }),
+    )
+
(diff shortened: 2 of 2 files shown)


parametrization:
Linking the verdict this close is based on for the record: Farhan Malik's ChangesRequested review — https://github.com/noorinalabs/noorinalabs-isnad-graph/pull/885#issuecomment-4454332820 — which correctly identified that user-service's `_build_post_login_url` always appends `/{provider}`, so the existing `auth/callback/:provider` route was already correct and this PR's change would have re-introduced #824. Closed as superseded-by-reality.

PR #914, opened 2026-05-17
fix(frontend): reconcile UserRole union with backend vocabulary (#876)

parametrization:
## Summary

Closes #876 (parent meta: noorinalabs/noorinalabs-main#161 — closed as tracked-in-children).

Reconciles the frontend `UserRole` union with the user-service backend vocabulary per **Option A** from the parent meta. The two vocabularies had overlapped on exactly one name (`admin`); every backend-only role name silently degraded to `'viewer'` via `deriveHighestRole`. A test assertion documented the silent-degrade as the "current correct" behaviour — this PR flips it.

**Before**: `UserRole = 'viewer' | 'editor' | 'moderator' | 'admin'` (frontend-invented in isolation; no product spec)
**After**: `UserRole = 'trial' | 'reader' | 'researcher' | 'admin'` (mirrors `ontology/repos/user-service.yaml:69` canonical hierarchy `{admin:40, researcher:30, reader:20, trial:10}`)

## Surface enumeration (all call sites audited)

Migrated (5 files):

| File | Lines | Change |
|------|-------|--------|
| `frontend/src/hooks/useAuth.ts` | 4 | `UserRole` union → backend vocabulary |
| `frontend/src/hooks/useAuth.ts` | 56-60 | `ROLE_HIERARCHY` table → `{trial:0, reader:1, researcher:2, admin:3}` |
| `frontend/src/hooks/useAuth.ts` | 64-67, 78 | Comment refresh + fallback `'viewer'` → `'trial'` (lowest tier, least permissive) |
| `frontend/src/hooks/__tests__/useAuth.test.ts` | (rewrite) | All assertions migrated; silent-degrade flipped to positive assertion (`['researcher','reader','trial'] → 'researcher'`); added "legacy names ignored" regression test |
| `frontend/src/hooks/__tests__/useSubscription.test.tsx` | 36 | Mock default `role: 'viewer'` → `'trial'` |
| `frontend/src/pages/admin/UserManagementPage.tsx` | 13 | `ROLES` const drives `<select>` options + `PATCH /api/v1/users/{id}/role` body |
| `frontend/src/pages/admin/UserManagementPage.tsx` | 23 | `useState` default `'viewer'` → `'trial'` |
| `frontend/src/pages/admin/UserManagementPage.tsx` | 171, 204 | Display fallbacks `?? 'viewer'` → `?? 'trial'` |
| `frontend/src/pages/ProfilePage.tsx` | 28-33 | `roleBadgeColor` switch cases migrated by tier-rank semantics |
| `frontend/src/pages/ProfilePage.tsx` | 179 | Display fallback `?? 'viewer'` → `?? 'trial'` |

Survives unchanged (admin-only checks — `'admin'` is in both vocabularies):

| File | Line | Why |
|------|------|-----|
| `frontend/src/components/AdminRoute.tsx` | 7 | `hasRole('admin')` |
| `frontend/src/components/AdminLayout.tsx` | 17, 23 | Uses `isAdmin` boolean derived from `roles.includes('admin')` |
| `frontend/src/components/Sidebar.tsx` | — | Only links to `/admin` route, no role gate |

## Semantics notes (no product input needed)

The frontend union was invented in isolation (per #161 wording: *"the frontend vocabulary is accidental divergence"*). There are no product-level equivalences between `viewer/editor/moderator` and `trial/reader/researcher` that needed translation — the names map cleanly by tier-rank within the new vocabulary:

- **`trial` → lowest tier / safe default** — replaces `'viewer'` as the fallback in `deriveHighestRole` and the default for `useState` and `?? 'viewer'` displays. `trial` is the least-permissive role, so an unknown-role user falls into the most-restricted bucket (better than silently being treated as a stand-in for any other tier).
- **`reader`** — mid-tier (rank 1). Was `editor` (rank 1) in the old hierarchy. ProfilePage badge color preserved as `--color-primary`.
- **`researcher`** — second-highest (rank 2). Was `moderator` (rank 2) in the old hierarchy. ProfilePage badge color preserved as `--color-warning`.
- **`admin`** — unchanged top tier (rank 3). Badge color unchanged (`--color-destructive`).

## Live-trace evidence

Post-fix grep of UserRole-context literal role strings in `frontend/src/`:

```
$ grep -rn "'viewer'\|'editor'\|'moderator'" frontend/src/
frontend/src/hooks/__tests__/useAuth.test.ts:34:    expect(deriveHighestRole(["viewer", "editor", "moderator"])).toBe("trial")
frontend/src/hooks/__tests__/useAuth.test.ts:38:    expect(deriveHighestRole(["editor", "admin"])).toBe("admin")
frontend/src/hooks/__tests__/useAuth.test.ts:39:    expect(deriveHighestRole(["viewer", "reader"])).toBe("reader")
```

Only three remaining references — all in **negative assertions** that prove the new code ignores the legacy vocabulary. No production code path returns or compares against the old names.

## Verification

```
$ npm run lint     # 0 errors, 11 warnings — all pre-existing on TimelinePage/ConfigPage, none from changed files
$ npm run test     # 12 files, 52 tests passed (2.79s)
$ npm run build    # tsc + vite — clean, no type errors (1492 modules transformed)
```

## Coordination for #877 (openapi-typescript codegen)

@Nneka — after this lands and you start #877, the lines to swap for generated `components['schemas']['UserRead']` (or whatever the user-service OpenAPI emits for the role type) are:

- `frontend/src/hooks/useAuth.ts:4` — `export type UserRole = 'trial' | 'reader' | 'researcher' | 'admin'`
- `frontend/src/hooks/useAuth.ts:56-60` — `ROLE_HIERARCHY` table (the rank values would still need to be hand-maintained, since OpenAPI doesn't model ordinal hierarchy — but the *member set* should come from codegen)

If user-service ships a typed enum schema (per Nadia's user-service#103 enum work — already merged), `openapi-typescript` should produce a string-literal union we can use directly here, deleting line 4 and importing the type.

## Tech Debt

None introduced. This PR closes an existing tech-debt item (#876, label `tech-debt`).

## Reviewers

- Primary: @Anya Kowalczyk (Tech Lead, isnad-graph)
- Secondary: @Ingrid Lindqvist (Senior Engineer, frontend, isnad-graph)

### DIFF
--- a/frontend/src/hooks/__tests__/useAuth.test.ts
+++ b/frontend/src/hooks/__tests__/useAuth.test.ts
@@ -7,34 +7,42 @@ describe("deriveHighestRole", () => {
     expect(deriveHighestRole(["admin"])).toBe("admin")
   })
 
-  it("returns 'viewer' when roles is empty", () => {
-    expect(deriveHighestRole([])).toBe("viewer")
+  it("returns 'trial' (lowest tier) when roles is empty", () => {
+    expect(deriveHighestRole([])).toBe("trial")
   })
 
-  it("returns the highest tier when multiple frontend-known roles are present", () => {
-    // moderator (rank 2) beats editor (rank 1)
-    expect(deriveHighestRole(["editor", "moderator"])).toBe("moderator")
+  it("returns the highest tier when multiple backend roles are present", () => {
+    // researcher (rank 2) beats reader (rank 1)
+    expect(deriveHighestRole(["reader", "researcher"])).toBe("researcher")
     // admin (rank 3) beats everything
-    expect(deriveHighestRole(["viewer", "editor", "moderator", "admin"])).toBe(
+    expect(deriveHighestRole(["trial", "reader", "researcher", "admin"])).toBe(
       "admin",
     )
   })
 
-  it("ignores backend role names that are not in the UserRole union", () => {
-    // user-service DB roles `researcher`, `reader`, `trial` aren't in
-    // the frontend UserRole union yet — they should be skipped.
-    expect(deriveHighestRole(["researcher", "reader", "trial"])).toBe("viewer")
+  it("returns the highest backend role and ignores unknown names", () => {
+    // Reconciled to backend vocabulary (#876): researcher/reader/trial are
+    // now first-class — no longer silently degrade to a stand-in tier.
+    expect(deriveHighestRole(["researcher", "reader", "trial"])).toBe(
+      "researcher",
+    )
+  })
+
+  it("ignores role names outside the backend vocabulary", () => {
+    // Legacy / unknown role names (e.g. the pre-#876 frontend-invented
+    // `viewer`/`editor`/`moderator`) are dropped; fall back to `trial`.
+    expect(deriveHighestRole(["viewer", "editor", "moderator"])).toBe("trial")
   })
 
   it("picks the highest matching tier even when unknown roles are mixed in", () => {
-    expect(deriveHighestRole(["researcher", "admin"])).toBe("admin")
-    expect(deriveHighestRole(["reader", "editor"])).toBe("editor")
+    expect(deriveHighestRole(["editor", "admin"])).toBe("admin")
+    expect(deriveHighestRole(["viewer", "reader"])).toBe("reader")
   })
 
   it("returns the single matching tier when only one known role is present", () => {
-    expect(deriveHighestRole(["editor"])).toBe("editor")
-    expect(deriveHighestRole(["moderator"])).toBe("moderator")
-    expect(deriveHighestRole(["viewer"])).toBe("viewer")
+    expect(deriveHighestRole(["reader"])).toBe("reader")
+    expect(deriveHighestRole(["researcher"])).toBe("researcher")
+    expect(deriveHighestRole(["trial"])).toBe("trial")
   })
 })
 
--- a/frontend/src/hooks/__tests__/useSubscription.test.tsx
+++ b/frontend/src/hooks/__tests__/useSubscription.test.tsx
@@ -33,7 +33,7 @@ function authAs(opts: { user: unknown; isAdmin: boolean }) {
     user: opts.user,
     loading: false,
     isAdmin: opts.isAdmin,
-    role: opts.isAdmin ? "admin" : "viewer",
+    role: opts.isAdmin ? "admin" : "trial",
     hasRole: () => opts.isAdmin,
     sessionExpired: false,
     isNewUser: false,
--- a/frontend/src/hooks/useAuth.ts
+++ b/frontend/src/hooks/useAuth.ts
@@ -1,7 +1,7 @@
 import { createContext, useContext, useState, useEffect, useCallback, type ReactNode } from 'react'
 import { createElement } from 'react'
 
-export type UserRole = 'viewer' | 'editor' | 'moderator' | 'admin'
+export type UserRole = 'trial' | 'reader' | 'researcher' | 'admin'
 
 /**
  * Shape of `/api/v1/users/me` (user-service UserRead schema).
@@ -50,33 +50,31 @@ interface AuthContextValue {
   refreshUser: () => Promise<void>
 }
 
+// Mirrors the user-service canonical hierarchy (ontology/repos/user-service.yaml:69
+// → { admin: 40, researcher: 30, reader: 20, trial: 10 }), rescaled to 0-indexed
+// ranks. Order matters for `deriveHighestRole` highest-first iteration.
 const ROLE_HIERARCHY: Record<UserRole, number> = {
-  viewer: 0,
-  editor: 1,
-  moderator: 2,
+  trial: 0,
+  reader: 1,
+  researcher: 2,
   admin: 3,
 }
 
 /**
- * Map a list of role names (as returned by user-service) to the highest
- * matching `UserRole` in the frontend hierarchy. Unknown role names
- * (e.g. user-service DB roles like `researcher`, `reader`, `trial` that
- * aren't yet represented in the frontend union) are ignored — callers
- * default to `viewer` for fully-unknown sets.
- *
- * NOTE: if `UserRole` is extended to cover additional backend roles, add
- * them to `ROLE_HIERARCHY` above and they will automatically participate
- * in derivation here (we iterate the hierarchy highest-first).
+ * Map a list of role names (as returned by user-service `UserRead.roles`) to
+ * the highest matching `UserRole`. The frontend union is reconciled with the
+ * backend vocabulary as of #876 — any string outside `trial|reader|researcher|admin`
+ * is unknown and ignored; if no known role matches, callers default to `trial`
+ * (the lowest tier — least permissive).
  */
 export function deriveHighestRole(roleNames: string[]): UserRole {
-  // Iterate ROLE_HIERARCHY from highest rank to lowest; return first match.
   const ordered = (Object.keys(ROLE_HIERARCHY) as UserRole[]).sort(
     (a, b) => ROLE_HIERARCHY[b] - ROLE_HIERARCHY[a],
   )
   for (const tier of ordered) {
     if (roleNames.includes(tier)) return tier
   }
-  return 'viewer'
+  return 'trial'
 }
 
 // Absolute URLs targeting the user-service vhost (`users.{base}`). Sourced
--- a/frontend/src/pages/ProfilePage.tsx
+++ b/frontend/src/pages/ProfilePage.tsx
@@ -27,9 +27,9 @@ function roleBadgeColor(role: string | null): string {
   switch (role) {
     case 'admin':
       return 'var(--color-destructive, #ef4444)'
-    case 'moderator':
+    case 'researcher':
       return 'var(--color-warning, #f59e0b)'
-    case 'editor':
+    case 'reader':
       return 'var(--color-primary)'
     default:
       return 'var(--color-muted-foreground)'
@@ -176,7 +176,7 @@ export default function ProfilePage() {
                 background: roleBadgeColor(profile.role),
               }}
             >
-              {(profile.role ?? 'viewer').toUpperCase()}
+              {(profile.role ?? 'trial').toUpperCase()}
             </span>
           </div>
         </div>
--- a/frontend/src/pages/admin/UserManagementPage.tsx
+++ b/frontend/src/pages/admin/UserManagementPage.tsx
@@ -10,7 +10,7 @@ import {
 } from '../../api/admin-client'
 import type { AdminUser } from '../../types/admin'
 
-const ROLES = ['viewer', 'editor', 'moderator', 'admin'] as const
+const ROLES = ['trial', 'reader', 'researcher', 'admin'] as const
 
 export default function UserManagementPage() {
   const queryClient = useQueryClient()
@@ -20,7 +20,7 @@ export default function UserManagementPage() {
   const [roleFilter, setRoleFilter] = useState('')
   const [statusFilter, setStatusFilter] = useState('')
   const [selectedIds, setSelectedIds] = useState<Set<string>>(new Set())
-  const [bulkRole, setBulkRole] = useState('viewer')
+  const [bulkRole, setBulkRole] = useState('trial')
   const [detailUserId, setDetailUserId] = useState<string | null>(null)
 
   const { data, isLoading, error } = useQuery({
@@ -168,7 +168,7 @@ export default function UserManagementPage() {
                   <td>{u.email}</td>
                   <td>{u.provider}</td>
                   <td>
-                    <select className="form-input" value={u.role ?? 'viewer'} onChange={(e) => roleMutation.mutate({ userId: u.id, role: e.target.value })} disabled={roleMutation.isPending} style={{ width: 120, padding: '0.25rem' }}>
+                    <select className="form-input" value={u.role ?? 'trial'} onChange={(e) => roleMutation.mutate({ userId: u.id, role: e.target.value })} disabled={roleMutation.isPending} style={{ width: 120, padding: '0.25rem' }}>
                       {ROLES.map((r) => <option key={r} value={r}>{r.charAt(0).toUpperCase() + r.slice(1)}</option>)}
                     </select>
                   </td>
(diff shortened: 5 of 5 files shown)


parametrization:
Requestor: Anya Kowalczyk
Requestee: Jun-Seo Park
RequestOrReplied: Approved
TechDebt: None

## Verdict

Reviewed at PR head_sha `671b09d516b161e6a5f80a8272b2a492627ef996` against the noorinalabs-isnad-graph frontend shape and the user-service backend canonical hierarchy. Approving.

## Acceptance checks (all run at head_sha via `gh api .../contents/<path>?ref=<sha>`)

1. **UserRole union flipped** — `useAuth.ts:4` reads `export type UserRole = 'trial' | 'reader' | 'researcher' | 'admin'`. PASS.
2. **ROLE_HIERARCHY values flipped** — `useAuth.ts:56-61` reads `{ trial: 0, reader: 1, researcher: 2, admin: 3 }`. Matches the PR body's claim and the user-service canonical `{admin:40, researcher:30, reader:20, trial:10}` ordering (rescaled to 0-indexed ranks — same ordinal positions). PASS.
3. **Silent-degrade flipped to positive** — `useAuth.test.ts:23-29` now asserts `deriveHighestRole(["researcher","reader","trial"]) === "researcher"`. The pre-#876 passing assertion that pinned the silent `'viewer'` fallback is gone. PASS.
4. **Legacy-name regression guard** — `useAuth.test.ts:31-35` asserts `deriveHighestRole(["viewer","editor","moderator"]) === "trial"`, proving the old names are dropped at the boundary rather than silently mapped. PASS.
5. **Production code sweep clean** — ran `grep -nE "[\"'](viewer|editor|moderator)[\"']"` across all `.ts`/`.tsx` files in `frontend/src/` excluding `__tests__/` at PR head_sha (62 files scanned). **Zero matches.** The 3 remaining occurrences in the repo are *exactly* the 3 negative-test-assertion lines in `useAuth.test.ts` (lines 34, 38, 39), all of which prove the new code ignores the legacy vocabulary. PR-body live-trace is exact. PASS.
6. **Survives-unchanged components verified** — fetched `AdminRoute.tsx`, `AdminLayout.tsx`, `Sidebar.tsx` at HEAD: `'admin'` literal still gates them; no legacy role literals appear. The admin-only contract is preserved because `admin` is the one role name common to both vocabularies. PASS.
7. **CI gates** — 7/8 jobs green at HEAD (lint-and-typecheck, test, frontend-lint-and-test, hooks-lint, scripts-lint, lockfile-validation, e2e-skipped). PASS for the gates that exercise this PR's surface. (Security-audit FAILURE — see Non-blocking note below.)

## Cross-references verified at canonical sources

- **Ontology citation** — `ontology/repos/user-service.yaml:69` reads `role_hierarchy: { admin: 40, researcher: 30, reader: 20, trial: 10 }`. PR's claim is exact, including the line number. PASS.
- **Issue #876** — body's Option-A recommendation is what was implemented; parent meta `noorinalabs-main#161` is referenced correctly as closed-tracked-in-children.
- **PR diff** — `gh pr view 914 --json files` reports `+44/-38` across the 5 claimed files; matches the surface-enumeration table in the PR body exactly.

## Tech-Lead angle review

**Vocabulary choice rationale (lossless tier-rank mapping)**

The PR makes the right call that no product input is needed:

- `viewer` → `trial` (rank 0 → rank 0): both are "default, least-permissive, unauthenticated-ish-tier" buckets. Using `trial` as the `?? 'viewer'` fallback also makes the *fail-closed direction* explicit — an unknown role now lands in the most-restricted bucket rather than silently being treated as a stand-in for a backend tier. This is the conservative move.
- `editor` → `reader` (rank 1 → rank 1): the old `editor` was a frontend invention with no product feature gated on it (zero call sites to `hasRole('editor')` in this repo per the surface enumeration). Mapping to `reader` preserves rank position without claiming any product equivalence.
- `moderator` → `researcher` (rank 2 → rank 2): same reasoning — no `hasRole('moderator')` consumer existed. Rank preserved.
- `admin` unchanged.

The "lossless" framing is accurate because the old names never had product semantics — they were just slots. The frontend-invented vocabulary had zero call-site discrimination beyond rank, so swapping names by rank is structurally sound and doesn't need a product sign-off. The audit table at the top of the PR body (10 changed locations across 5 files; 3 surviving locations across 3 files) is exhaustive — I cross-checked the survives-unchanged list at head_sha and they're correct.

**Pre-existing 11 warnings in TimelinePage/ConfigPage — out-of-scope is correct**

These weren't in the changed files at HEAD. Folding unrelated lint cleanup into a vocabulary-reconcile PR would muddy the diff and the live-trace. Filing as a follow-on if warning-noise becomes pain. No change requested.

**`roleBadgeColor` switch in `ProfilePage.tsx` — design-system color semantics**

Read at HEAD (`ProfilePage.tsx:26-37`):

```ts
case 'admin':      return 'var(--color-destructive, #ef4444)'
case 'researcher': return 'var(--color-warning, #f59e0b)'
case 'reader':     return 'var(--color-primary)'
default:           return 'var(--color-muted-foreground)'  // trial falls through
```

Tier-rank semantics are preserved (admin=destructive/red = highest-privilege-warning, researcher=warning/amber, reader=primary/blue, trial=muted/grey-default). This matches the *previous* color-by-rank mapping (admin=destructive, moderator=warning, editor=primary, viewer=muted), so the visual identity is stable across the migration — users with the same backend rank will see the same badge color before and after. The explicit hex fallbacks (`#ef4444`, `#f59e0b`) inside the `var()` are kept as defense for environments where the design-system tokens haven't loaded yet, which matches the broader convention in this file (e.g., `roleBadgeColor` for `admin` already had `#ef4444` as a defensive fallback pre-PR). No change requested.

**`trial` as the unknown-role fallback — fail-closed direction**

This is the highest-quality decision in the PR. Pre-#876, `deriveHighestRole` of an unknown set returned `'viewer'` *which was a real frontend role* — meaning unknown-role users got the same UI treatment as legitimate viewers. Post-#876, the fallback is `'trial'`, which is now also the explicit "you have no real role yet" tier — so unknown-role users get the most-restricted UI, which is the correct fail-closed direction for an auth-gated frontend. The comment at `useAuth.ts:67` ("if no known role matches, callers default to `trial` (the lowest tier — least permissive)") captures the *why* clearly enough that I don't need to ask for additional documentation.

## Sequencing coordination note (Nneka's #877 codegen)

The PR body's call-out to Nneka — `useAuth.ts:4` (union) and `useAuth.ts:56-60` (ROLE_HIERARCHY) — is accurate at HEAD. When openapi-typescript codegen lands in #877:

- Line 4 can be deleted and the type imported from `components['schemas']['UserRoleName']` (or whatever name user-service#103 emits — the merged user-service enum work should produce a string-literal union).
- Lines 56-60 stay hand-maintained: OpenAPI doesn't model ordinal hierarchy, only the member set. The *order/rank* is frontend application logic. Best practice would be a runtime assertion at module init that `Object.keys(ROLE_HIERARCHY).sort() === [...generated_union].sort()` to catch backend-side additions automatically. That's a #877-scope concern, not blocking here.

Recommend Nneka cite this PR's `useAuth.ts:4` and `:56-60` line numbers verbatim in #877's spawn brief to anchor the swap site.

## Non-blocking note (not introduced by this PR — identical to IG#912)

CI `security-audit` shows FAILURE on this PR's run (run 25980899915). Root cause is *the same* `urllib3 2.6.3` advisory pair (CVE-2026-44431, CVE-2026-44432) flagged on IG#912 — it's the backend `uv.lock` baseline, not anything in this frontend-only PR. The 7 frontend-relevant gates pass. **Not blocking IG#914.** This will hit *every* isnad-graph PR until the dep bump lands, so the W11 dep-cluster issue I flagged from IG#912 is also the resolution path here.

## Throughline-watch

Two W11 patterns worth a one-liner in retro:

1. **Vocabulary-reconcile PRs as a class** — IG#914 (frontend↔backend role vocab) follows the same shape as the user-service `UserRole` enum work (user-service#103, already merged). Both close a "we made up names in isolation" gap by adopting the canonical source's vocabulary. If we see a third instance (e.g., an internal-tooling repo coining its own subscription-tier names), this is a recurring tech-debt class that argues for an ontology-tier "name reconciliation audit" — a periodic check that string-literal unions/enums across repos agree on the canonical name set for shared domain concepts.
2. **Universal CI noise from security-audit urllib3** — IG#912 *and* IG#914 both have the same security-audit FAILURE from `urllib3 2.6.3`, on completely unrelated diffs. That's now 2 of 2 W11 isnad-graph PRs hitting it. A bump-urllib3 PR landing at the head of the W11 isnad-graph cluster would unblock the entire wave's PR-state cleanliness; suggest treating it as a wave-blocker for `security-audit` PR-status hygiene even though no individual PR is blocked by it.

No surface drift beyond the declared scope. Approving.

parametrization:
Requestor: Ingrid Lindqvist
Requestee: Jun-Seo Park
RequestOrReplied: Approved
TechDebt: ProfilePage `roleBadgeColor` renders an inline-styled `<span>` rather than the design-system `Badge` component (which already has `destructive`/`default`/`outline`/`secondary` variants). This is pre-existing tech debt, NOT introduced by this PR — I confirmed by reading the diff at PR head_sha — but the W11 vocab migration would have been a natural moment to also lift it onto `Badge`. Flagging as latent design-system-consumer-debt for a follow-up issue, not as a blocker on this PR.

## Senior-Engineer-frontend review (UserManagementPage, ProfilePage, role-gated rendering)

Verified at head_sha `671b09d516b161e6a5f80a8272b2a492627ef996`.

### 1. Migration completeness (the throughline-watch)

Cloned PR branch, ran a recursive grep on `frontend/src/` for the OLD vocab `viewer|editor|moderator`:

```
frontend/src/hooks/__tests__/useAuth.test.ts:33:    // `viewer`/`editor`/`moderator`) are dropped; fall back to `trial`.
frontend/src/hooks/__tests__/useAuth.test.ts:34:    expect(deriveHighestRole(["viewer", "editor", "moderator"])).toBe("trial")
frontend/src/hooks/__tests__/useAuth.test.ts:38:    expect(deriveHighestRole(["editor", "admin"])).toBe("admin")
frontend/src/hooks/__tests__/useAuth.test.ts:39:    expect(deriveHighestRole(["viewer", "reader"])).toBe("reader")
```

All 4 remaining occurrences are **intentional test cases asserting the old vocab is now correctly ignored**. No orphaned production references. Migration is exhaustive.

### 2. `deriveHighestRole` semantics flip — positive assertion, no silent degrade

The previous behaviour fell through to `'viewer'` when no role matched; the new behaviour falls through to `'trial'` (the lowest tier — least permissive). The unit test `it("returns 'trial' (lowest tier) when roles is empty")` makes this contract explicit, and the prior `it("returns the highest backend role and ignores unknown names")` flips from a silent-degrade pattern to a positive assertion that `researcher` wins.

This is a strictly safer default — least-privilege fall-through. No breakage risk for downstream gates.

### 3. `useAuth` ROLE_HIERARCHY rank values — match the user-service ontology rescale

`useAuth.ts:52-58` comment cites `ontology/repos/user-service.yaml:69` for `{ admin: 40, researcher: 30, reader: 20, trial: 10 }` and correctly rescales to the 0-indexed `{ trial: 0, reader: 1, researcher: 2, admin: 3 }`. Citation is verifiable; rescale preserves ordering. `hasRole(minRole)` continues to do `rank[userRole] >= rank[minRole]` so gating semantics are unchanged — only the labels move.

### 4. UserManagementPage UX — the admin role-picker workflow

`ROLES = ['trial', 'reader', 'researcher', 'admin'] as const` drives:
- The role-filter dropdown (top of table)
- The bulk-role `<select>` (when rows are selected)
- The per-row inline role `<select>` in the Role column
- The bulk-role action default-state (`useState('trial')`)
- The per-row null-fallback (`u.role ?? 'trial'`) and detail-modal fallback (`userDetail.role ?? 'trial'`)

The cascade `trial → reader → researcher → admin` makes UX sense as an admin role-picker: it's an ascending-privilege list, which is the conventional read for role-management UIs (left-to-right or top-to-bottom in dropdown = lower-to-higher privilege). Default state `bulkRole: 'trial'` is the least-disruptive default (admin clicks "Set Role" on a multi-select bulk action by accident → demotes to lowest-tier rather than promotes to admin), which is a safer accident-floor than the previous `'viewer'` default for the same reason (least-privilege).

One UX nuance worth noting (NOT a blocker): the per-row dropdown defaults to `u.role ?? 'trial'` for users with no role. If a user genuinely has no role assigned in the backend, the dropdown will VISUALLY display `Trial` even though no `trial` role exists on their record. If an admin then clicks `Save` (or any other side-effect), this would write `trial` to the backend rather than preserve null. I don't see a Save button per-row — role changes appear to fire immediately via `onChange → roleMutation.mutate` — so the moment the admin even tabs through a "no role" user the dropdown would persist `trial`. Worth a follow-up issue if this matters, but the pre-PR behaviour was identical (defaulted to `'viewer'` and would persist that), so this is pre-existing semantic and NOT a regression. Skipping a blocker on this.

### 5. ProfilePage `roleBadgeColor` cascade — visual hierarchy

```
admin     → var(--color-destructive) (red/vermillion)
researcher → var(--color-warning)    (yellow/ochre)
reader    → var(--color-primary)     (sienna/blue depending on theme)
default   → var(--color-muted-foreground) (grey — covers trial + null + unknown)
```

The cascade reads sensibly:
- **admin = destructive** — high-attention/high-privilege, common pattern (matches GitHub's "Owner" red badge).
- **researcher = warning** — middle-tier elevated, distinct from primary.
- **reader = primary** — neutral "active user" tone.
- **default (trial / null / unknown) = muted-foreground** — de-emphasized, signaling "no special privilege" or "unverified".

Aligns with design-system semantic-color intent (`primary` = brand, `warning` = caution-worthy elevated, `destructive` = high-impact / irreversible / superuser). I'd consider it a minor improvement over silent default-color-everything.

That said: as noted in the TechDebt line above, ProfilePage uses an inline-styled `<span>` rather than the design-system `Badge` component. The `Badge` component has `default`/`destructive`/`secondary`/`outline` (and brand-specific `sunni`/`shia`/`sahih`) but no `warning` variant — so a follow-up that lifts onto `Badge` would also need a `warning` variant added to design-system or fall back to `outline`. That's a forward-looking observation for a separate W12-style issue, not this PR's responsibility.

### 6. Admin-adjacent surface audit (the explicit ask)

Read at PR head_sha:
- `AdminRoute.tsx` — gates on `hasRole('admin')` (literal). Unaffected by vocab change because `'admin'` is in both old and new vocab.
- `AdminLayout.tsx` — gates on `isAdmin` which derives from `roles.includes('admin')`. Same conclusion.
- `Sidebar.tsx` — conditionally renders Admin nav-link on `isAdmin`. Same.
- `useSubscription.ts` — `enabled: !!user && !isAdmin` (admin bypass). Same.
- `App.tsx:81` — comment-only reference to "Admin routes — require isAdmin".

**No `'moderator'`-gated admin-adjacent surfaces were missed.** Pre-PR, the `moderator` tier existed in the union but no production code actually gated on it (only `hasRole('admin')` / `isAdmin` patterns). The vocab change is safe to land without touching the gating components.

### 7. Test count verification

Recursive `grep -c '^\s+it(' src/**/*.test.{ts,tsx}` at PR head returns:
- 12 distinct test files
- 52 total `it()` blocks

Matches the brief's "12 files / 52 tests pass" claim. The new useAuth.test.ts adds 11 cases (3 of which are the deliberate negative assertions for old vocab being ignored). The useSubscription.test.tsx 1-line diff is a mock update flipping `role: 'viewer'` → `role: 'trial'` to match the new union.

(I did NOT run the test suite locally — frontend dev depends on a local design-system tarball at `/tmp/ig-pr914/noorinalabs-design-system-0.0.1.tgz` which is a monorepo wiring detail. The test files themselves are deterministic unit tests over a pure vocab-mapping function; I verified them by reading the assertions at PR head.)

### 8. Throughline-watch (design-system + isnad-graph dual-context)

I'm also reviewing DS#85 (container tokens) right now, so I have design-system context warm. One throughline observation **for the team-lead, NOT a blocker on this PR**:

This PR migrates UserRole TypeScript-level. The display layer (ProfilePage's badge `<span>`, UserManagementPage's `<select>`) is hand-styled with inline CSS-var lookups, not lifted onto design-system components. The DS#85 container-token PR I'm also reviewing reinforces the pattern that design-system is the canonical home for visual semantics. A future "lift role display onto design-system Badge" issue would close that gap and would also need:
- Adding a `warning` variant to `Badge` (for the researcher tier), OR
- Mapping researcher → existing `outline` variant.

Flagging only — DS team can pick up.

### Verdict

Approved. Clean, scoped reconciliation; surface exhaustively swept; safer default fall-through; tests cleanly assert both positive and negative semantics; admin-gating surfaces unaffected. One pre-existing tech-debt item flagged (ProfilePage badge → design-system Badge) but explicitly NOT introduced by this PR.

Co-reviewer: Anya Kowalczyk (primary, isnad-graph frontend architecture authority).

Issue #421, opened 2026-03-28
feat(frontend): auth-gated routing and protected routes


parametrization:
## Description
Wrap all application routes in auth protection so unauthenticated users are redirected to the login page. Currently App.tsx routes directly to content with zero auth checks.

## Acceptance Criteria
- [ ] `ProtectedRoute` wrapper component checking `useAuth()`, redirects to `/login` if unauthenticated
- [ ] `AdminRoute` wrapper checking `isAdmin`, shows 403 or redirects if not admin
- [ ] All main app routes wrapped with `ProtectedRoute` in App.tsx
- [ ] All `/admin/*` routes wrapped with `AdminRoute`
- [ ] `/login` and `/auth/callback/:provider` are public (not protected)
- [ ] Loading spinner while auth state is initializing (no redirect flash)
- [ ] Return URL preserved — after login, redirect to originally requested page (store in query param or state)

## Dependencies
- Depends on login page (issue for `/login` route) and OAuth callback (issue for `/auth/callback`) being at least stubbed so protected routes can redirect to them.
- Can develop in parallel, integrate at end.

Filed by: Fatima Okonkwo

PR #991, opened 2026-06-12
fix(frontend): guard non-JSON auth responses in LoginPage (#977)

parametrization:
Closes #977

## Context
Defensive frontend complement to **deploy#420 / PR #421** (prod-login JSON.parse). The deploy-side fix promotes the stale prod image + adds an anti-drift smoke guard; this PR is the **code guard** so a future origin/runtime-config drift degrades gracefully instead of throwing a cryptic parse error.

## Problem
`LoginPage.tsx` called `res.json()` on the auth endpoints with no content-type / parse guard on the success paths. When the user-service origin mis-resolves — or any misroute serves the SPA `index.html` with **HTTP 200 + text/html** — `res.json()` parses `<!DOCTYPE html…` and throws a raw `Unexpected token '<'` at the user. The 200 status masks the misroute.

## Fix
- New helper `frontend/src/lib/authJson.ts` → `readJsonResponse<T>(res)`: checks `Content-Type` is `application/json` (charset-suffix tolerant) and guards the parse, throwing `NonJsonResponseError` with a clean user-facing message (`NON_JSON_RESPONSE_MESSAGE`) instead of the raw `SyntaxError`.
- Routed **every** auth-flow `res.json()` in `LoginPage` through it:
  - providers probe (`/auth/providers`) — already fell back to defaults; now consistent
  - OAuth init (`/auth/oauth/{provider}/login`) — success + error paths
  - email login (`/auth/login/email`) — success + error paths
  - register (`/auth/register`) — success + error paths
- Error paths keep their existing `.catch(() => ({ detail: HTTP <status> }))` fallback.

Scoped to LoginPage + the new shared helper. The runtime-config mechanism itself is untouched (deploy-side).

## Test
- `src/lib/__tests__/authJson.test.ts` (6): genuine JSON, charset-suffixed JSON, 200 `text/html` misroute → `NonJsonResponseError`, absent content-type, JSON-typed-but-unparseable body, and "never leaks the raw `Unexpected token` SyntaxError".
- `src/pages/__tests__/LoginPage.test.tsx` (2): HTML-misrouted email login surfaces the clean error and persists **no** `access_token`; a genuine JSON login still succeeds.

## Test Plan
- [x] `vitest run` (new files) — 8/8 pass
- [x] `eslint` clean on changed files
- [x] `tsc -b` (build typecheck) clean
- [ ] Reviewers: @Nneka Obi (primary), @Anya Kowalczyk (secondary)

### DIFF
--- a/frontend/src/lib/__tests__/authJson.test.ts
+++ b/frontend/src/lib/__tests__/authJson.test.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect } from 'vitest'
+
+import {
+  readJsonResponse,
+  NonJsonResponseError,
+  NON_JSON_RESPONSE_MESSAGE,
+} from '../authJson'
+
+describe('readJsonResponse', () => {
+  it('parses a genuine application/json response', async () => {
+    const res = new Response(JSON.stringify({ access_token: 'abc' }), {
+      status: 200,
+      headers: { 'Content-Type': 'application/json' },
+    })
+    await expect(readJsonResponse<{ access_token: string }>(res)).resolves.toEqual({
+      access_token: 'abc',
+    })
+  })
+
+  it('honours a charset-suffixed JSON content-type', async () => {
+    const res = new Response(JSON.stringify([{ a: 1 }]), {
+      status: 200,
+      headers: { 'Content-Type': 'application/json; charset=utf-8' },
+    })
+    await expect(readJsonResponse(res)).resolves.toEqual([{ a: 1 }])
+  })
+
+  it('throws NonJsonResponseError on a 200 text/html body (the deploy#420 misroute)', async () => {
+    // A reverse-proxy fallback serving the SPA index.html with HTTP 200.
+    const res = new Response('<!DOCTYPE html><html><body>SPA</body></html>', {
+      status: 200,
+      headers: { 'Content-Type': 'text/html' },
+    })
+    await expect(readJsonResponse(res)).rejects.toBeInstanceOf(NonJsonResponseError)
+    await expect(readJsonResponse(res)).rejects.toThrow(NON_JSON_RESPONSE_MESSAGE)
+  })
+
+  it('throws NonJsonResponseError when content-type is absent', async () => {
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Anya Kowalczyk
Requestee: Idris Yusuf
RequestOrReplied: Approved

Secondary (tech-lead) review of #991 — verified via diff + files API at head `I.Yusuf/0977-loginpage-json-guard` and CI (all checks green; e2e skipping by design).

**Abstraction quality (a).** `readJsonResponse<T>()` is a clean, reusable guard — content-type check is charset-suffix tolerant (`includes('application/json')` after lowercasing), the parse is wrapped so a JSON-typed-but-malformed body also degrades, and the generic keeps call sites typed. The error type carries its own default message and a stable name. Nicely minimal, no over-engineering.

**Coverage (b).** Confirmed zero bare `res.json()` left on any auth path in LoginPage — providers probe, OAuth init (success + error), email login (success + error), and register (success + error) all route through the helper. Error paths keep their existing `.catch(() => detail HTTP <status>)` fallback, so a misroute on an error response still yields a sane message rather than the helper's throw escaping.

**Message hygiene (c).** `NON_JSON_RESPONSE_MESSAGE` is user-appropriate — it names the auth service as unreachable/misconfigured and suggests retry, leaking no internals. The raw `SyntaxError` never reaches `setError`, and a dedicated test asserts the alert never renders the "unexpected token" string. Good.

**useAuth scope (d).** Agreed the unguarded `res.json()` calls in `useAuth.ts` (token-refresh, sessions list, `/me`) are correctly OUT of scope here — the issue is LoginPage-only. They are real, though: same misroute class (`window.RUNTIME_CONFIG` drift → SPA index.html on 200) would throw the same cryptic parse error on those post-auth flows. A low-priority follow-up to extend `readJsonResponse` into `useAuth` is warranted — recommend filing it so the helper becomes the project-wide convention rather than a LoginPage one-off.

No blockers. Approved.

Issue #1000, opened 2026-06-13
enable DS color tokens as Tailwind @theme utilities (bg-card/text-foreground/border-border resolve) — prerequisite for #967 color-utility migration


parametrization:
## Problem

The isnad-graph frontend's Tailwind build generates **zero** atomic color/background utility rules. Verified against the built CSS bundle (`dist/assets/index-*.css`), not by reasoning:

- `bg-card`, `bg-background`, `bg-primary`, `bg-accent`, `bg-muted`, `border-border`, `text-foreground`, `text-muted-foreground` → **0 generated rules**. They are **no-ops** in the shipped app today.
- The only color classes that resolve are DS **component** classes (`.text-success`, `.text-danger`, `.text-warning`, `.feature-card`, `.stat-card`, `.flex-row`, `.muted-text`, …) and Tailwind-native **non-color** utilities (spacing `p-4`/`gap-2`, radius `rounded-lg`, layout `flex`/`grid`).

### Latent example
`frontend/src/pages/AuthCallbackPage.tsx` error card:
```tsx
<div className="... rounded-2xl border border-border bg-card p-8 shadow-xl ...">
```
`bg-card` and `border-border` contribute **no background and no border** — the card renders without them (the inner icon's color comes from an inline `var(--color-destructive)`, not the class).

### Root cause
Per **DS #104**, `@noorinalabs/design-system` ships its color tokens to `:root` as plain CSS (so any non-Tailwind consumer gets them), **not** into Tailwind's `@theme` namespace. Tailwind 4 only generates `bg-*`/`text-*`/`border-*` color utilities from colors declared in `@theme`, so it never learns `card`/`foreground`/`border`/etc. as colors. (Spacing/radius coincidentally work because the DS scale is numerically identical to Tailwind's built-in scale.)

## Why this matters
This blocks the **color half** of the #967 inline-style → DS-utility alignment intent. #981 (bucket D) could only migrate static *layout/spacing/radius* props to utilities and DS *component* classes; color-bearing inline styles had to stay inline because there is no working color-utility target. To finish the alignment, the app must expose the DS color tokens as Tailwind theme colors.

## Proposed fix (needs design + DS-owner coordination)
[...]

PR #1001, opened 2026-06-13
refactor(frontend): migrate ad-hoc inline styles to DS/Tailwind utility composition (#981)

parametrization:
Closes #981 — #967 bucket D, step 3 of the FE color/style chain (#979 → #980 → #981). Conservative, provably-non-regressing migration of the heaviest 3 files off ad-hoc inline `style={{}}` objects toward Tailwind-native utility classes + DS component-class composition. **Pattern** migration only — same rendered output, no value changes.

## What migrated (static styling → utilities / DS classes)
- **Layout / spacing / radius / display / flex / grid / font-weight → Tailwind utilities.** Verified 1:1: the DS spacing scale is numerically identical to Tailwind's (both 4px base), radius utilities resolve to `var(--radius-*)`, and font-weight tokens equal the numerics (`--font-weight-semibold:600` etc., so `font-semibold`==`fontWeight:600`).
- **Whole elements → existing DS component classes** where the DS rule matches property-for-property: `.muted-text` (`color:var(--color-muted-foreground)`) for the many pure muted-color spans, `.small-muted` for a `text-sm`+muted pair.
- Net **−189 lines** across the 3 files.

## What stayed inline (intentional — documented)
- **Color / background / border-color.** This app's Tailwind build generates **zero** atomic semantic-color utilities — `bg-card`/`text-foreground`/`border-border` are **no-ops** in the shipped bundle (the DS ships tokens to `:root`, not Tailwind `@theme`). Converting color tokens to those classes would **regress**, so they stay inline. The enabler is filed as **#1000** (prerequisite for the color half of #967).
- **font-size.** Tailwind `text-*` forces a paired `line-height` the inline `fontSize` did not set → would change vertical rhythm. Kept inline.
- **box-shadow / font-family / transition** (Tailwind maps these to its own scales, not the DS tokens), **dynamic graph-overlay positions** (legend `right: detailOpen ? 340 : 12`, tooltip/zoom coords), and **conditional values** (active-state backgrounds).
- **Physical margins** (`ml-2`/`mt-0.5`) used rather than logical (`ms`/`me`) to guarantee **no RTL behavior change** in this no-regression pass; logical-property migration is a separate concern.

## Files
- `frontend/src/pages/GraphExplorerPage.tsx` (78 → 42 inline-style objects)
- `frontend/src/pages/PricingPage.tsx` (26 → 21)
- `frontend/src/pages/admin/UserManagementPage.tsx` (11 → 6; note: 11, not the issue's stale "26")

The residual inline styles are exclusively the kept-inline categories above (verified: no leftover `display`/`gap`/`padding`/`flex`/`radius`/`textTransform`/`cursor`/`overflow` props remain inside any `style={{}}`).

## Test Plan — no-regression verification
Since the full app stack can't run in the sandbox for a browser pixel-diff, non-regression is proven by **computed-CSS parity** instead of eyeballing:
- **Every** utility class introduced (91 distinct) was verified to **generate a real CSS rule in the built bundle** — none is a silent no-op (the exact failure mode that makes `bg-card` inert). Spacing/radius/font-weight values equal the DS tokens they replaced.
- `npm run build` (tsc + vite) — green.
- `npm run lint` — 0 errors (1 pre-existing `set-state-in-effect` warning at GraphExplorerPage:65, in untouched data-merge code).
- `npm run test` — **186/186 pass**, including the 13 GraphExplorerPage tests.

## Reviewers
@Ravi Desai (primary, DS-composition lens), @Junseo Park (secondary, code lens)

Refs #967. Follow-up: #1000 (wire DS color tokens into Tailwind `@theme`).

### DIFF
--- a/frontend/src/pages/GraphExplorerPage.tsx
+++ b/frontend/src/pages/GraphExplorerPage.tsx
@@ -184,13 +184,7 @@ export default function GraphExplorerPage() {
   return (
     <div className="flex flex-col" style={{ height: 'calc(100vh - 64px)' }}>
       {/* --- TOOLBAR --- */}
-      <div
-        className="flex flex-wrap items-center gap-3 px-4 py-2"
-        style={{
-          borderBottom: '1px solid var(--color-border)',
-          background: 'var(--color-card)',
-        }}
-      >
+      <div className="flex flex-wrap items-center gap-3 border-b border-border bg-card px-4 py-2">
         {/* Search */}
         <div className="relative w-[300px]">
           <input
@@ -212,11 +206,10 @@ export default function GraphExplorerPage() {
                 setSearchInput('')
                 setSearchOpen(false)
               }}
-              className="absolute right-2 top-1/2 -translate-y-1/2 cursor-pointer border-none"
+              className="absolute right-2 top-1/2 -translate-y-1/2 cursor-pointer border-none text-muted-foreground"
               style={{
                 background: 'none',
                 fontSize: '1rem',
-                color: 'var(--color-muted-foreground)',
               }}
               aria-label="Clear search"
             >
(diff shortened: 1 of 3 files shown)


parametrization:
Re-closing to match the locked plan: superseded by a fresh PR2 (rebased on the merged #1002 bridge) that carries the full #981 migration — the layout/DS-class work here plus color/background/border-color -> atomic utilities. (The brief reopen above crossed with that decision.) #981 stays open; PR2 will Close it.

parametrization:
Reopened as the #981 vehicle, expanded in place — head 8c897e4 carries the full migration (layout + color→atomic utilities), rebased on the merged bridge f3fffb4. The duplicate fresh PR (#1003) is closed in favor of this. Final state; not toggling again.

PR #1063, opened 2026-06-14
feat(admin): configurable log retention (keep last X days) knob (#1038)

### LINKED ISSUE #1038: feat(admin): configurable log retention — 'keep last X days' knob in admin Config panel
## Request (owner, 2026-06-14)

> "We have limited space on the VPS so we need some kind of log cleanup/rotation and only keep the last X number of days (I should be able to configure this in the admin panel)."

## Current state (verified)

The retention *mechanism* mostly exists, but the **value is hardcoded** and **not admin-configurable**:
- **Loki** (aggregated app logs): `retention_period: 168h` (7d) + compactor `retention_enabled: true` in `infra/loki/loki-config.yml` — auto-deletes past 7d, but changing it requires a redeploy.
- **Prometheus**: `--storage.tsdb.retention.time=30d`.
- **Docker container logs**: per-service `json-file` caps (see deploy#450 for the 9 uncapped services).

What's missing is the **admin-panel knob** to set "keep last X days" without a redeploy.

## Design (recommended)

The admin Config surface already exists and is DB-backed: `src/api/routes/admin/config.py` (`GET`/`PATCH` `SystemConfig`, with `/audit`) + `frontend/src/pages/admin/ConfigPage.tsx`. Add a **`log_retention_days`** field to `SystemConfig`:
- **Backend**: extend `SystemConfig` model + `update_config` validation (min/max bounds, e.g. 1–365); persists to the existing config table; audited via the existing config-audit path.
- **Frontend**: add a numeric "Log retention (days)" control to `ConfigPage.tsx` via `admin-client.ts`.
- **Enforcement** (cross-repo → deploy): the persisted value must actually drive Loki retention. Loki's static `retention_period` isn't hot-reloadable, but its **`runtime_config`** per-tenant `retention_period` override IS. Wiring tracked in **deploy#451** (this issue depends on it).

This is the "profile.json"-style idea generalized: one DB-backed setting, surfaced in the panel, instead of a config-file redeploy.

## Acceptance
- [ ] `log_retention_days` settable in admin Config panel (bounded), persisted + audited
- [ ] Value drives effective Loki retention (via deploy#451)
- [ ] Non-admins cannot change it (AdminRoute / RBAC)

## Provenance
New intake 2026-06-14. Depends on deploy#451. Recommend **P5W3** (not P5W2 "API light-up"). Companion hygiene: deploy#450.

parametrization:
## Summary

- Adds a configurable **"Log retention (days)"** knob (1–365) to the admin Config panel (ig#1038) so an admin can cap log history on the disk-constrained VPS **without a redeploy**.
- Persists + audits `log_retention_days` through the existing `system_config` / `config_audit` path; bounds enforced by pydantic (`422` out of range).
- Propagates the value to Loki's hot-reloadable `runtime_config` overrides on the API side (the deploy#451 contract), so the new window takes effect with no Loki restart.

**isnad-graph-only** — the deploy half (deploy#451 / noorinalabs-deploy#455) is already merged. No deploy-repo changes here.

## Related Issues

Closes #1038

## What shipped

**Backend**
- `SystemConfig` / `SystemConfigUpdate` gain `log_retention_days` — `Field(default=7, ge=1, le=365)`. Persisted + audited via the same path as every other config field.
- New `LokiSettings` (`env_prefix=LOKI_`, `runtime_overrides_path` default `/etc/loki/runtime/overrides.yaml`).
- New `src/api/loki_retention.py` — best-effort **in-place truncate-write** of Loki's runtime-overrides file (preserves the inode the `loki_runtime` volume tracks; surgical rewrite of only the `retention_period` line). Logged **no-op** when the volume/file is absent (local/dev/test); never raises — the DB write stays source of truth.
- `update_config` calls the writer **only when `log_retention_days` changes**.

**Frontend**
- `ConfigPage` gains a bounded numeric "Keep last N days of logs" control (1–365), wired into the existing PATCH save flow, styled consistently with the sibling controls.

## Enforcement contract (drives Loki via deploy#451)

The persisted `log_retention_days` is the value deploy#451's Loki `runtime_config` override consumes:

1. Admin PATCHes `log_retention_days` → persisted to Postgres (source of truth) + audited.
2. The API rewrites `overrides.<tenant>.retention_period` in `/etc/loki/runtime/overrides.yaml` (the `loki_runtime` volume, mounted **rw into api**, **read-side into loki**) as `{days*24}h`.
3. Loki re-reads `runtime_config` every 30s and applies the new `retention_period` on its next compaction — **no restart**.

Verified against the merged deploy contract (deploy#455): path `/etc/loki/runtime/overrides.yaml`; named volume `loki_runtime` shared rw between `loki-runtime-init` (seed), `api` (write), `loki` (read); tenant id `fake`; static fallback `limits_config.retention_period = 168h` (7d) when no override is present. The volume is seeded once and **not** clobbered on redeploy, so an admin-set value persists.

**RBAC:** `PATCH /api/v1/admin/config` is gated by `require_admin` (unchanged) — non-admins cannot read or set the control.

## Test plan

- `tests/test_api/test_loki_retention.py` (pure, no DB) — duration conversion, in-place rewrite with **inode preservation**, structure preservation, no-op safety (dir/file absent, key missing → not corrupted). **6/6 pass locally.**
- `tests/test_api/test_admin_config.py::TestLogRetention` — round-trip persist+read, bounds rejection (422), RBAC rejection, propagation-on-change, no-write-on-unrelated-change. DB-backed `TestClient` fixture (hangs without a live DB in the sandbox) → **covered by CI**.
- `frontend/.../ConfigPage.test.tsx` — control renders from config + PATCHes the changed value. **2/2 pass locally.**
- Local gates: `ruff check` / `ruff format --check` (src+tests) ✓, `mypy src/` ✓, frontend `eslint` (0 errors) ✓, `tsc --noEmit` ✓.

Acceptance:
- [x] `log_retention_days` settable in admin Config panel (bounded 1–365), persisted + audited
- [x] Effective-retention contract documented (drives Loki via deploy#451)
- [x] Non-admins cannot change it (RBAC)

## Review Checklist
- [ ] CI checks passing (lint, typecheck, test, security-audit)
- [ ] Peer reviewed by another engineer
- [ ] Must-fix items resolved
- [ ] Tech debt items filed as GitHub Issues (if any)
- [ ] Issues closed after merge

## Must-Fix Items
None

## Tech Debt Items
None

### DIFF
--- a/frontend/src/pages/admin/ConfigPage.tsx
+++ b/frontend/src/pages/admin/ConfigPage.tsx
@@ -9,6 +9,7 @@ interface SystemConfig {
   feature_flags: Record<string, boolean>
   max_search_results: number
   max_pagination_limit: number
+  log_retention_days: number
 }
 
 interface ConfigAuditEntry {
@@ -74,6 +75,7 @@ export default function ConfigPage() {
         rate_limit_per_minute: config.rate_limit_per_minute,
         max_search_results: config.max_search_results,
         max_pagination_limit: config.max_pagination_limit,
+        log_retention_days: config.log_retention_days,
         cors_origins: [...config.cors_origins],
         feature_flags: { ...config.feature_flags },
       })
@@ -104,6 +106,9 @@ export default function ConfigPage() {
     }
     if (formData.max_pagination_limit !== config?.max_pagination_limit) {
       update.max_pagination_limit = formData.max_pagination_limit
+    }
+    if (formData.log_retention_days !== config?.log_retention_days) {
+      update.log_retention_days = formData.log_retention_days
     }
     const newOrigins = corsInput.split(',').map((s) => s.trim()).filter(Boolean)
     if (JSON.stringify(newOrigins) !== JSON.stringify(config?.cors_origins)) {
@@ -208,6 +213,26 @@ export default function ConfigPage() {
             className="form-input-block"
           />
         </label>
+      </section>
+
+      <section className="section-mb">
+        <h2>Log Retention</h2>
+        <label>
+          Keep last N days of logs
+          <input
+            type="number"
+            min={1}
+            max={365}
+            value={formData.log_retention_days ?? 7}
+            onChange={(e) =>
+              setFormData((prev) => ({ ...prev, log_retention_days: Number(e.target.value) }))
+            }
+            className="form-input-block"
+          />
+        </label>
+        <p className="muted-text">
+          Application logs older than this are rotated/deleted (1–365 days). Frees disk on the VPS.
+        </p>
       </section>
 
       <section className="section-mb">
--- a/frontend/src/pages/admin/__tests__/ConfigPage.test.tsx
+++ b/frontend/src/pages/admin/__tests__/ConfigPage.test.tsx
@@ -1,0 +1,80 @@
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+
+import ConfigPage from '../ConfigPage'
+
+const CONFIG = {
+  rate_limit_per_minute: 60,
+  cors_origins: ['http://localhost:3000'],
+  feature_flags: {},
+  max_search_results: 100,
+  max_pagination_limit: 100,
+  log_retention_days: 7,
+}
+
+function jsonResponse(body: unknown, status = 200): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { 'Content-Type': 'application/json' },
+  })
+}
+
+function renderPage() {
+  const queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false } },
+  })
+  return render(
+    <QueryClientProvider client={queryClient}>
+      <ConfigPage />
+    </QueryClientProvider>,
+  )
+}
+
+describe('ConfigPage — log retention (ig#1038)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
+  it('renders the log-retention control populated from config', async () => {
+    vi.stubGlobal(
+      'fetch',
+      vi.fn(() => Promise.resolve(jsonResponse({ ...CONFIG, entries: [], total: 0 }))),
+    )
+    renderPage()
+
+    await screen.findByText('Log Retention')
+    const input = screen.getByLabelText(/Keep last N days/i) as HTMLInputElement
+    expect(input.value).toBe('7')
+  })
+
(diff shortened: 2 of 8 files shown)


parametrization:
Requestor: Marisol Vega-Cruz
Requestee: Nneka Obi
RequestOrReplied: Approved
TechDebt: loki writer fires whenever the field is present in the PATCH body (mirrors the always-write audit row) rather than strictly on value-change, and the overrides rewrite is count=1 single-tenant — both fine for the current deploy#451 seed, worth a forward note if per-tenant retention ever diverges
Must-fix: none

Reviewed the diff at HEAD plus the full update_config route and models at the PR head ref. Backend correctness and the config/audit/RBAC contract all hold up.

**Generic persistence + audit (no plumbing skipped).** log_retention_days flows through the same exclude_none path every sibling field uses. body.model_dump(exclude_none=True) only surfaces it when set; it is checked against FORBIDDEN_CONFIG_KEYS, upserted into system_config, and appended to config_audit by the same generic loop — so old/new values are captured exactly like rate_limit_per_minute. No per-field branch, no audit gap. On read-back, _load_config sees the int default of 7 so isinstance(default, int) routes through int(raw); the round-trip stays an int (test_round_trip confirms).

**Validation.** Field(default=7, ge=1, le=365) on both SystemConfig and SystemConfigUpdate gives a 422 on 0 and on 366 (both asserted). Server-side bounds are the real gate; the HTML min/max on the control are advisory only, which matches how the other numeric controls on this page already behave.

**RBAC.** PATCH stays gated by require_admin; test_non_admin_cannot_set overrides the dep to raise and asserts the non-admin PATCH is rejected. (GET /config has no admin gate, but that is pre-existing surface the new field merely rides — out of scope here.)

**Writer invocation + isolation.** apply_loki_retention runs after the DB writes, only when the key is present in updates (test_propagates / test_no_loki_write_for_unrelated_change both pass). It never raises — every failure path (dir absent, file absent, key missing, OSError) returns False and logs — and its return is intentionally ignored, so a file no-op or write failure cannot disturb the already-persisted DB value, which stays the source of truth. day->hour conversion is days*24 with an h suffix (7=168h, 30=720h, 1=24h, verified), and the int() cast on an already-bounded Pydantic value is sound.

**File-writer safety (cross-check of Jelani's lens).** Confirmed the in-place truncate write (open(path, "w")) preserves the inode the loki volume view tracks — test_rewrites_value_in_place asserts st_ino is unchanged — and the surgical regex rewrites only retention_period while leaving the header/tenant/indentation intact, and is idempotent (the .*$ consumes the prior trailing comment so it does not accumulate). The key-missing path leaves the file byte-identical rather than half-written.

**Frontend.** The bounded numeric control (1-365) is populated from config, diffed against config in the save handler, and PATCHed only when changed via the admin client — same shape as max_search_results / max_pagination_limit. Tests cover the populated render and the change->PATCH body.

Clean, well-scoped, fully additive, and CI is green across all eight checks. Approving.

PR #1099, opened 2026-06-17
tech-debt(grades): single source of truth for GRADE_LABELS (backend → frontend) (#1054)

parametrization:
## What

Establishes a **single source of truth** for the hadith-grade vocabulary so the backend and frontend copies of `GRADE_LABELS` can no longer drift (ig#1054, follow-up TechDebt from PR #1052 review).

Before: `GRADE_LABELS` (canonical token → display label) was hand-maintained in **both** `src/utils/grades.py` and `frontend/src/lib/grades.ts`.

## Approach

Backend is the source of truth; the frontend **derives** from it via a committed generated artifact, gated against drift — mirroring the repo's existing `gen:types` drift-gate idiom and the `emit_ingest_schema.py` emit/check pattern.

- **`scripts/emit_grade_vocab.py`** (`emit` / `check`) reads `GRADE_LABELS` from `src.utils.grades` and writes the ordered map to **`frontend/src/lib/grade-vocab.generated.json`**.
- **`frontend/src/lib/grades.ts`** imports that JSON instead of a literal map. `GRADE_TOKENS` (derived) and the colour/bar helpers — which are presentation, not vocabulary — stay frontend-side. Display/facet key order is preserved.
- **Drift gate (can't-drift enforcement):**
  - CI: new `Check grade-vocab drift` step in `ci.yml`'s `lint-and-typecheck` job (runs where the Python source is importable, not the node-only frontend job).
  - Mirrored by a `grade-vocab-drift` **pre-commit** hook (full local⇄CI parity).
  - `emit()` guards the backend invariant `set(GRADE_LABELS) == set(GRADE_TOKENS)`.
  - Backend tests assert the invariant + that the committed JSON matches `GRADE_LABELS`.

## Verification (local, all green)

- Backend: `emit_grade_vocab.py check` in sync; ruff check/format; mypy; `pytest tests/test_utils/test_grades.py` 43 passed.
- Frontend: `tsc --noEmit` clean; `eslint src/` 0 errors; **full vitest 323 passed**.
- pre-commit ⇄ CI sync-drift gate: OK.

## Acceptance

- [x] A single source of truth for grade tokens/labels; frontend derives rather than duplicates.

Closes #1054

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,14 @@ jobs:
       - run: uv run ruff check src/ tests/
       - run: uv run ruff format --check src/ tests/
       - run: uv run mypy src/
+      # Grade-vocab single-source-of-truth gate (ig#1054). The frontend's
+      # GRADE_LABELS is generated from src/utils/grades.py into
+      # frontend/src/lib/grade-vocab.generated.json; fail if the committed JSON
+      # has drifted from the backend. Mirrored by the `grade-vocab-drift`
+      # pre-commit hook. Runs here (not in the node-only frontend job) because it
+      # needs to import the Python source of truth.
+      - name: Check grade-vocab drift (backend <-> frontend)
+        run: uv run python scripts/emit_grade_vocab.py check
 
   security-audit:
     runs-on: ubuntu-latest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -105,6 +105,19 @@ repos:
       # Integration tests require Docker (Neo4j, PostgreSQL) — run in CI only.
       # To run locally: make infra && ENVIRONMENT=test uv run pytest tests/integration/ -m integration
 
+      # Grade-vocab single-source-of-truth gate (ig#1054). Mirrors the
+      # "Check grade-vocab drift" step in ci.yml's lint-and-typecheck job: fails
+      # if frontend/src/lib/grade-vocab.generated.json has drifted from the
+      # backend GRADE_LABELS. Regenerate with `uv run python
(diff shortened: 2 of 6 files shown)


parametrization:
Requestor: Anya Kowalczyk
Requestee: Ingrid Lindqvist
RequestOrReplied: Approved

Reviewed the diff at HEAD. This establishes a genuine single source of truth and a drift gate that actually bites.

The backend `src/utils/grades.py` GRADE_LABELS becomes the only hand-authored copy; `scripts/emit_grade_vocab.py emit` renders it to `frontend/src/lib/grade-vocab.generated.json`, and `grades.ts` now imports that JSON instead of re-declaring the map. The colour/bar helpers correctly stay in the frontend since they are presentation, not vocabulary.

The gate catches drift in both directions because `check` regenerates the expected JSON in memory from the backend and byte-compares it against the committed file, failing on any mismatch or a missing file. It is wired in two enforced places — the `Check grade-vocab drift` step in ci.yml's lint-and-typecheck job (Python, so it can import the source of truth) and the `grade-vocab-drift` pre-commit hook whose `files` filter covers all three relevant paths (grades.py, the emit script, the generated JSON), so a stage of any of them retriggers it. `_render` also guards the backend invariant that GRADE_LABELS keys equal GRADE_TOKENS, refusing to emit a vocabulary with a missing or stray label.

Display order is preserved and tested — `json.dumps(dict(GRADE_LABELS))` keeps insertion order, and `test_frontend_grade_vocab_in_sync_with_backend` asserts both equality and key-order equality, with a comment noting the frontend facet derives display order from Object.keys. The sync-drift gate, cspell, actionlint, and full test suite are all green and the PR is mergeable. Approving.

PR #1153, opened 2026-07-01
feat(graph): ṭabaqa-based generation layering in the isnad-tree explorer

parametrization:
## Summary

Adds **ṭabaqa-based generation layering** to the isnad-tree graph explorer (Phase-7 criterion #5, closes #1043). A new **Ṭabaqa** layout mode arranges narrator nodes into chronological generation bands (top→bottom, earliest first) so the transmission tree reads chronologically, with a canvas overlay annotating the bands.

## What changed

- **`frontend/src/lib/tabaqa.ts`** (new) — pure layering logic:
  - `normalizeTabaqa` maps a raw narrator `generation` to a canonical ṭabaqa key, mirroring the backend `NarratorGeneration` enum (`sahabi → tabii → taba_tabii → atba_taba_tabiin → later`) plus common spelling/transliteration aliases (e.g. `Companion`, `tabi_tabiin`).
  - `computeTabaqaLayers` groups nodes into contiguous bands ordered by generation, sorting within each band chronologically by death/birth year (undated last). Narrators with **unknown/unmapped generation collect into a trailing band** so every node still renders (graceful degradation).
  - `tabaqaRankByNode` flattens layers into a node→band-rank map for the renderer.
- **`ForceGraph.tsx`** — in `tabaqa` mode, pins each node's `fy` to its generation band while leaving `x` force-driven, so narrators spread within their chronological row.
- **`GraphExplorerPage.tsx`** — Ṭabaqa layout toggle + a band-legend overlay listing the generations in chronological order.
- **i18n** — `layoutTabaqa` + ṭabaqa band labels added across all seven locales (parity-test compliant).
- **`ontology/structural/`** — regenerated (structural-ontology staleness gate).

## Tests (fail on pre-fix)

- `frontend/src/lib/__tests__/tabaqa.test.ts` — layering logic: canonical ordering, contiguous ranks, unknown-band trailing placement, chronological intra-band sort, alias normalization.
- `frontend/src/pages/__tests__/GraphExplorerPage.test.tsx` — wiring: the Ṭabaqa toggle passes `layoutMode` + per-band ranks to the graph (unknown generation ranked last); the band overlay renders in chronological order.

Both suites are coupled to symbols/UI that do not exist on the base branch, so they fail pre-fix.

## Verification

- `npx vitest run` — 355 passed
- `npm run build` (tsc + vite) — green
- `npx eslint src/` — 0 errors
- Full pre-commit (incl. pytest-unit, pip-audit, structural-ontology, eslint, typecheck) — passed; pre-push frontend-build — passed

Reviewers: @Ravi Wickramasinghe (peer, UX), @Anya Kowalczyk (secondary, tech lead).


parametrization:
## UX / visualization review — Ravi (peer, ig#1043)

Reviewed the artifact at HEAD `f0ca40a` (diff + `contents?ref=`), no parent checkout. Solid, well-tested feature — clean pure layering module, robust alias normalization, machine-enforced locale parity. One interaction regression on the existing toggles keeps this from a clean approve.

### What works well
- **Chronology direction is correct.** `fy = (rank − (layerCount−1)/2)·90` puts rank 0 (`sahabi`) at the most-negative y = top of canvas, later generations descending — bands read top→bottom, earliest first. The band-legend box renders `tabaqaLayers` in rank order, so the legend list order matches the on-canvas row order. Clear and correctly ordered.
- **Graceful degradation is genuinely graceful.** Unknown/unmapped generations collect into the trailing `unknown` band, only non-empty bands emit rows (contiguous ranks), and the rank map covers every node — so nothing is dropped or left unpinned, and the catch-all sits at the bottom row. Switching *away* from ṭabaqa recreates nodes without `fy`, correctly releasing the pins (no stuck rows).
- **i18n is airtight.** All 8 keys land in all 7 locales, the label-key table matches the JSON keys exactly, and `locales.test.ts` mechanically enforces same-base-keys-as-`en` plus no-empty-strings. `ar`/`ur` carry proper Arabic-script labels for RTL.

### Requested change (blocking, minor severity, ~2 lines)
- **Toggling the still-cosmetic modes now reshuffles the graph.** On the base branch `ForceGraph` never received the layout prop, so `hierarchy`/`radial` were pure button-highlight no-ops. This PR adds the layout prop to the `graphData` `useMemo` dependency list, and since the band-rank map is `null` and the band count is constant when not in ṭabaqa mode, toggling among `force`/`hierarchy`/`radial` now recreates the entire node array → the force sim re-ingests and reheats → a visible reshuffle. That's a regression: it makes those three modes *look* like they did something when they still don't, which is more confusing than the prior honest no-op.
  - Suggested fix: collapse the non-ṭabaqa modes for the memo, e.g. derive `pinMode = layoutMode === 'tabaqa' ? 'tabaqa' : 'free'` and depend on `pinMode` instead of the raw layout string. Only entering/leaving ṭabaqa reheats; the cosmetic toggles stay no-ops until they gain real behavior.

### Non-blocking suggestions (fold into the revision if cheap — not filed as tech-debt)
- The intra-band chronological sort has no *visual* effect: every node in a band shares one `fy` and x is force-driven, so `sortChronologically` only sets Map insertion order. The PR body/tests imply within-band chronology, but there's no left→right ordering inside a row. Fine to keep (vertical banding carries the chronology) — worth a one-line comment noting the sort is for determinism, not layout, so a future reader doesn't expect a horizontal effect.
- There are no on-canvas band dividers/row labels — the only band annotation is the corner legend box. An MVP-acceptable choice, but on-canvas row labels would make the chronology land without cross-referencing the legend.
- In `en`, "Later followers" (`atba_taba_tabiin`) vs "Later generations" (`later`) both open with "Later" — mildly ambiguous side by side.

Happy to re-review promptly once the toggle memo is scoped.

---
Requestor: Ravi Wickramasinghe
Requestee: Nneka Obi
RequestOrReplied: ChangesRequested
TechDebt: none

